PR #16, opened 2024-04-16
Switch to ADTypes.jl v1.0

gdalle:
- Define `TracerSparsityDetector` following the `AbstractSparsityDetector` framework of ADTypes v1.0
- Remove SparseDiffTools extension
- Add tests
- Add docs
- Commit manifests and tweak docs CI because ADTypes v1.0 is not released
- Put test extras in the main project file to avoid https://github.com/JuliaLang/Pkg.jl/issues/1585


gdalle:
As you wish, no rush

PR #33, opened 2024-05-03
Remove AbstractIndexSet typing

gdalle:
Remove constraint `S <: AbstractSet{<:Integer}` in case we want to use vectors or other stuff.

Try to document the expectation of the tracer with respect to the "set" type.

See https://discourse.julialang.org/t/best-data-structure-for-fast-unions-of-large-sets-of-integers/113785/

### DIFF
--- a/src/tracers.jl
+++ b/src/tracers.jl
@@ -1,7 +1,5 @@
 abstract type AbstractTracer <: Number end
 
-const AbstractIndexSet = AbstractSet{<:Integer}
-
 # Convenience constructor for empty tracers
 empty(tracer::T) where {T<:AbstractTracer} = empty(T)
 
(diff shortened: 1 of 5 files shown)


src/tracers.jl
```diff
@@ -13,19 +11,24 @@ empty(tracer::T) where {T<:AbstractTracer} = empty(T)
     ConnectivityTracer{S}(indexset) <: Number
 
 Number type keeping track of input indices of previous computations.
-The provided index set type `S` has to be an `AbstractSet{<:Integer}`, e.g. `BitSet` or `Set{UInt64}`.
+The provided index set type `S` has to satisfy the following conditions:
+
+- it is an iterable with `<:Integer` element type
+- it implements methods `union` and `push!`
```
adrhill, line 17: > it implements methods `union` and `push!`

I just checked: `union`, `union!` and `push!` are required

adrhill (changes requested):

src/tracers.jl
gdalle: why `union!`? when do you ever mutate a `Tracer` (which is essentially a number) in-place?

adrhill: For the `HessianTracer`, I take a `deepcopy` of one of the dicts and mutate it in place. It doesn't have to be this way, but it currently is.

adrhill: As mentioned in #34, the `HessianTracer` leaves room for optimization.

gdalle: How about this custom dict merging?

```julia
julia> d1 = Dict(:a => Set([1, 2, 3]), :b => Set([3, 4, 5]))
Dict{Symbol, Set{Int64}} with 2 entries:
  :a => Set([2, 3, 1])
  :b => Set([5, 4, 3])

julia> d2 = Dict(:b => Set([5, 6, 7]), :c => Set([8, 9, 10]))
Dict{Symbol, Set{Int64}} with 2 entries:
  :b => Set([5, 6, 7])
  :c => Set([10, 9, 8])

julia> mergewith(union, d1, d2)
Dict{Symbol, Set{Int64}} with 3 entries:
  :a => Set([2, 3, 1])
  :b => Set([5, 4, 6, 7, 3])
  :c => Set([10, 9, 8])
```

gdalle: See #35 for this discussion, I have added `union!` in the meantime

gdalle:
I think this is ready, but I would rename all the occurrences of the set type to `S`. At the moment there is a mix of `S` and `T`. Thoughts?

adrhill:
Yes, this is a good idea. `T` for tracer types, `S` for set types.

src/tracers.jl
adrhill: Commented on in #35. Your proposed `mergewith` already exists and is called `additive_merge`. `union!` is used as inside the current implementation of `additive_merge`.

gdalle:
I have put `S` in the few places I had spotted `T`, there might be more out there but I suggest we merge

adrhill (approved):

PR #4, opened 2024-04-01
Compare with Symbolics

gdalle:
- Add a comparison to the test suite for the `brusselator` example
- Add a benchmarking file to play around, and the results are convincing!

```julia
julia> benchmark_brusselator(6, :tracer)
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  128.471 μs …   5.457 ms  ┊ GC (min … max):  0.00% … 94.62%
 Time  (median):     143.145 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   170.058 μs ± 313.636 μs  ┊ GC (mean ± σ):  12.42% ±  6.53%

  ▃▅▇█▇▆▅▄▃▂▂▁                                                  ▂
  █████████████▇█▇▆▆▆▆▅▄▆▄▅▄▅▄▃▄▇▆▅▆▄▆▆▆▆▅▇▆▆▅▅▄▃▄▄▄▄▁▁▁▄▃▅▁▅▄▆ █
  128 μs        Histogram: log(frequency) by time        361 μs <

 Memory estimate: 261.39 KiB, allocs estimate: 2408.

julia> benchmark_brusselator(6, :symbolics)
BenchmarkTools.Trial: 1219 samples with 1 evaluation.
 Range (min … max):  3.390 ms … 17.434 ms  ┊ GC (min … max): 0.00% … 68.25%
 Time  (median):     3.662 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   4.097 ms ±  1.471 ms  ┊ GC (mean ± σ):  7.88% ± 13.68%

  ▃██▆▄▂▁▁                                                    
  ████████▇█▇▇▅▆▆▄▇▆▅▅▅▅▄▅▅▅▅▄▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄▅▇▇▇▇█▆▄▄ █
  3.39 ms      Histogram: log(frequency) by time     9.84 ms <

 Memory estimate: 2.86 MiB, allocs estimate: 41727.

julia> benchmark_brusselator(24, :tracer)
BenchmarkTools.Trial: 1411 samples with 1 evaluation.
 Range (min … max):  2.379 ms … 14.906 ms  ┊ GC (min … max):  0.00% … 74.13%
 Time  (median):     2.952 ms              ┊ GC (median):     0.00%
[...]

PR #38, opened 2024-05-03
Add `SortedVector` set type

gdalle:
Add a new pseudoset type `SortedVector` which is uniformly faster than ~~both `BitSet` and~~ `Set` up to 10% sparsity (see [Discourse](https://discourse.julialang.org/t/best-data-structure-for-fast-unions-of-large-sets-of-integers/113785/32?u=gdalle))


test/sortedvector.jl
```diff
@@ -21,3 +23,7 @@ using Test
         end
     end
 end;
+
+sd = TracerSparsityDetector(SortedVector{UInt, Vector{UInt}})
+@test ADTypes.jacobian_sparsity(diff, rand(10), sd) isa SparseMatrixCSC
+@test_broken ADTypes.hessian_sparsity(x -> sum(abs2, diff(x)), rand(10), sd) isa SparseMatrixCSC
```
adrhill, line 29: Why would this be broken?

src/sortedvector.jl
```diff
@@ -0,0 +1,78 @@
...
+
+function SortedVector(data::V; already_sorted=false) where {T,V<:AbstractVector{T}}
+    return SortedVector{T,V}(data; already_sorted)
+end
+
+Base.show(io::IO, sv::SortedVector) = print(io, "SortedVector($(sv.data))")
+
+function Base.union(v1::SortedVector{T,V}, v2::SortedVector{T,V}) where {T,V}
+    left, right = v1.data, v2.data
+    both = similar(left, length(left) + length(right))
```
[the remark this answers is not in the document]
gdalle: I like `both` because it sits in the middle between `left` and `right`. And `union` is absurd, we're literally inside the `union` function

This is performance-independent bikeshedding, we can do that after NeurIPS

src/sortedvector.jl
```diff
@@ -0,0 +1,78 @@
...
+    # common part of left and right
+    @inbounds while (
+        left_index in eachindex(left) &&
+        right_index in eachindex(right) &&
+        both_index in eachindex(both)
+    )
+        left_item = left[left_index]
+        right_item = right[right_index]
+        left_smaller = left_item <= right_item
+        right_smaller = right_item <= left_item
```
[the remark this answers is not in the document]
gdalle: Because they are both used in the `ifelse`s underneath.
We want to increment `left_index` if `left_smaller` and `right_index` if `right_smaller`, but if `left_item == right_item` we can have _both_ `left_smaller` and `right_smaller`. This is important to get rid of duplicates

src/sortedvector.jl
```diff
@@ -0,0 +1,78 @@
...
+    @inbounds while left_index in eachindex(left) && both_index in eachindex(both)
+        both[both_index] = left[left_index]
+        left_index += 1
+        both_index += 1
+    end
+    @inbounds while right_index in eachindex(right) && both_index in eachindex(both)
+        both[both_index] = right[right_index]
+        right_index += 1
+        both_index += 1
+    end
```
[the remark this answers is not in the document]
gdalle: Marginal performance gain, possibly a loss with a view, I'd say ship it first and worry later

test/sortedvector.jl
gdalle: Because it requires `copy!`, see #36

gdalle: See https://github.com/adrhill/SparseConnectivityTracer.jl/issues/39

src/sortedvector.jl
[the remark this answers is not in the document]
adrhill: The name `both` confused me while reviewing and I have to maintain this code.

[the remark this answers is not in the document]
gdalle: if it's confusing and not just esthetic that's different

replaced it with `result`

src/sortedvector.jl
```diff
@@ -0,0 +1,86 @@
...
+        if already_sorted
+            new{T,V}(data)
+        else
+            new{T,V}(sort(data))
+        end
+    end
+
+    function SortedVector{T,V}(x::Number) where {T,V<:AbstractVector{T}}
+        return new{T,V}(convert(V, [T(x)]))
+    end
```
gdalle, line 35: is there a faster way?

src/sortedvector.jl
```diff
@@ -0,0 +1,86 @@
...
+
+    function SortedVector{T,V}() where {T,V<:AbstractVector{T}}
+        return new{T,V}(convert(V, T[]))
+    end
```
gdalle, line 39: could this be improved by referencing a constant empty `SortedVector` like you did with other types?

src/sortedvector.jl
adrhill: I'm not aware of one.

src/sortedvector.jl
adrhill: Couldn't we expect `new{T,V}(V())` to work?

adrhill:
Out of curiosity: Do we need to update #4 again? 😉

src/sortedvector.jl
gdalle: you're right!

gdalle:
> Out of curiosity: Do we need to update https://github.com/adrhill/SparseConnectivityTracer.jl/pull/4 again? 😉

Oh definitely. They've been blown so far out of the water they don't even see the water

src/sortedvector.jl
gdalle: let's keep it for now

adrhill:
In `runtests.jl`, there are a couple of loops over different set types:

https://github.com/adrhill/SparseConnectivityTracer.jl/blob/83cb0b07ea4973e95a9031c9468af50e0927d679/test/runtests.jl#L46-L48

https://github.com/adrhill/SparseConnectivityTracer.jl/blob/83cb0b07ea4973e95a9031c9468af50e0927d679/test/runtests.jl#L259-L262

`SortedVector` should be tested there as well.

And once `HessianTracer` is supported via `distributive_merge`, the same applies to `HessianTracer`:

https://github.com/adrhill/SparseConnectivityTracer.jl/blob/83cb0b07ea4973e95a9031c9468af50e0927d679/test/runtests.jl#L118-L120

Once this passes, we make `SortedVector` the `DEFAULT_SET_TYPE` in `src/pattern.jl`.

gdalle:
> Once this passes, we make SortedVector the DEFAULT_SET_TYPE in src/pattern.jl.

Keep in mind that `BitSet` is still faster for denser cases. Should we be asymptotically optimal and let the user tweak for smaller sizes? I think so but it's your call

gdalle:
Should I get rid of the parametric vector type and just use `Vector` for simplicity?

adrhill:
> Keep in mind that BitSet is still faster for denser cases. Should we be asymptotically optimal and let the user tweak for smaller sizes?

The benchmarks on Discourse were primarily on huge input dimensionalities (`10e6`), whereas I feel users interested in sparsity probably deal with smaller problems. I think the to select the default, we should benchmark on inputs of dimensionality 10 to 1000. Maybe there is a clear winner regardless of sparsity.

adrhill:
Closes #41.

Issue #39, opened 2024-05-03
Get rid of `union!` and `push!`


gdalle:
Conceptually, tracers are like numbers, so they should not be modified in place.
In particular, when implementing a new pseudo-set type like `SortedVector` in #38, one should only need to define `union`

adrhill:
Feel free to address this in #38 by modifying `distributive_merge`.

PR #42, opened 2024-05-03
Get rid of union!

gdalle:
Modify `promote_order` and `distributive_merge` to use `reduce`

PR #44, opened 2024-05-03
Merge `HessianTracer`s without `union!`

gdalle:


adrhill:
Closes #39 after #42 was reverted.

Issue #40, opened 2024-05-03
HessianTracer type improvements


gdalle:
https://github.com/adrhill/SparseConnectivityTracer.jl/blob/83cb0b07ea4973e95a9031c9468af50e0927d679/src/tracers.jl#L119-L121

Would it be more coherent to have keys of type `eltype(S)`, since they are also indices, rather than always `UInt64`?

adrhill:
Sounds good. :)

gdalle:
As seen in #45, we should also

- make the `AbstractDict` subtype parametric, e.g. to use `SortedDict` in conjunction with `SortedVector`
- rename the type parameter `T` to `I` for integer wherever it still appears
- add an inner constructor that checks `eltype(S) == I`

adrhill:
See also #51.

PR #45, opened 2024-05-03
Improve `HessianTracer` performance

adrhill:
* Couple key-type of `HessianTracer` dict to set-type
* Move sorting kwarg into inner constructor of `SortedVector`
* When merging `HessianTracer`s, make a set out of the keys outside of the inner loop. This should avoid repeated sorting for `union`s with `SortedVector`.

### DIFF
--- a/src/pattern.jl
+++ b/src/pattern.jl
@@ -194,8 +194,8 @@ function hessian_pattern(f, x, ::Type{S}=DEFAULT_SET_TYPE) where {S}
 end
 
 function hessian_pattern_to_mat(
-    xt::AbstractArray{HessianTracer{S}}, yt::HessianTracer{S}
-) where {S}
+    xt::AbstractArray{HessianTracer{S,T}}, yt::HessianTracer{S,T}
+) where {S,T}
     # Allocate Hessian matrix
     n = length(xt)
     I = UInt64[] # row indices
--- a/src/sortedvector.jl
+++ b/src/sortedvector.jl
@@ -5,13 +5,13 @@ A wrapper for sorted vectors, designed for fast unions.
 
 # Constructor
 
-    SortedVector(data::AbstractVector; already_sorted=false)
+    SortedVector(data::AbstractVector; sorted=false)
 
 # Example
 
 ```jldoctest
 x = SortedVector([3, 4, 2])
-x = SortedVector([1, 3, 5]; already_sorted=true)
+x = SortedVector([1, 3, 5]; sorted=true)
 z = union(x, y)
 
 # output
@@ -22,8 +22,9 @@ SortedVector([1, 2, 3, 4, 5])
 struct SortedVector{T<:Number} <: AbstractVector{T}
     data::Vector{T}
 
-    function SortedVector{T}(data::AbstractVector{T}) where {T}
-        return new{T}(convert(Vector{T}, data))
+    function SortedVector{T}(data::AbstractVector{T}; sorted=false) where {T}
+        sorted_data = ifelse(sorted, data, sort(data))
+        return new{T}(convert(Vector{T}, sorted_data))
     end
 
     function SortedVector{T}(x::Number) where {T}
@@ -35,13 +36,8 @@ struct SortedVector{T<:Number} <: AbstractVector{T}
     end
 end
 
-function SortedVector(data::AbstractVector{T}; already_sorted=false) where {T}
-    sorted_data = ifelse(already_sorted, data, sort(data))
-    return SortedVector{T}(sorted_data)
-end
-
 function Base.convert(::Type{SortedVector{T}}, v::Vector{T}) where {T}
-    return SortedVector(v; already_sorted=false)
+    return SortedVector{T}(v; sorted=false)
 end
 
 Base.eltype(::SortedVector{T}) where {T} = T
@@ -82,5 +78,5 @@ function Base.union(v1::SortedVector{T}, v2::SortedVector{T}) where {T}
         result_index += 1
     end
     resize!(result, result_index - 1)
-    return SortedVector(result; already_sorted=true)
+    return SortedVector{T}(result; sorted=true)
 end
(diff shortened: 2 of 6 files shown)


src/tracers.jl
```diff
@@ -168,15 +194,18 @@ function additive_merge(a::HessianTracer, b::HessianTracer)
...
+    sb = keys2set(S, db)
+
+    # add second-order interaction term by ignoring symmetry
     for (ka, va) in pairs(da)
-        da[ka] = union(va, keys(db))
+        da[ka] = union(va, sb)
     end
     for (kb, vb) in pairs(db)
-        db[kb] = union(vb, keys(da))
+        db[kb] = union(vb, sa)
```
adrhill, line 208: These are the most crucial changes to avoid repeated sorting when using `SortedVector`.

src/sortedvector.jl
```diff
@@ -5,13 +5,13 @@ A wrapper for sorted vectors, designed for fast unions.
 
 # Constructor
 
-    SortedVector(data::AbstractVector; already_sorted=false)
+    SortedVector(data::AbstractVector; sorted=false)
```
gdalle, line 8: I liked `already_sorted` because it clearly refers to the input, whereas `sorted=false` feels very wrong for a `SortedVector`

src/sortedvector.jl
```diff
@@ -22,8 +22,9 @@ SortedVector([1, 2, 3, 4, 5])
 struct SortedVector{T<:Number} <: AbstractVector{T}
     data::Vector{T}
 
-    function SortedVector{T}(data::AbstractVector{T}) where {T}
-        return new{T}(convert(Vector{T}, data))
+    function SortedVector{T}(data::AbstractVector{T}; sorted=false) where {T}
+        sorted_data = ifelse(sorted, data, sort(data))
+        return new{T}(convert(Vector{T}, sorted_data))
     end
```
gdalle, line 28: we could also keep the non-parametrized version of this constructor

src/pattern.jl
```diff
@@ -194,8 +194,8 @@ function hessian_pattern(f, x, ::Type{S}=DEFAULT_SET_TYPE) where {S}
 end
 
 function hessian_pattern_to_mat(
-    xt::AbstractArray{HessianTracer{S}}, yt::HessianTracer{S}
-) where {S}
+    xt::AbstractArray{HessianTracer{S,T}}, yt::HessianTracer{S,T}
+) where {S,T}
```
gdalle, line 198:
```suggestion
    xt::AbstractArray{HessianTracer{S,I}}, yt::HessianTracer{S,I}
) where {S,I}
```

src/tracers.jl
```diff
@@ -120,8 +132,8 @@ $SET_TYPE_MESSAGE
 
 For a higher-level interface, refer to [`hessian_pattern`](@ref).
 """
-struct HessianTracer{S} <: AbstractTracer
-    inputs::Dict{UInt64,S}
+struct HessianTracer{S,I<:Integer} <: AbstractTracer
+    inputs::Dict{I,S}
```
gdalle, line 136: add an inner constructor that checks `eltype(S) == I`

src/tracers.jl
```diff
@@ -120,8 +132,8 @@ $SET_TYPE_MESSAGE
 
 For a higher-level interface, refer to [`hessian_pattern`](@ref).
 """
-struct HessianTracer{S} <: AbstractTracer
-    inputs::Dict{UInt64,S}
+struct HessianTracer{S,I<:Integer} <: AbstractTracer
```
gdalle, line 135: parametrize Hessian by the dict type in order to allow `SortedDict`s (we can assume it's `<:AbstractDict`)

gdalle, line 135:
```suggestion
struct HessianTracer{I<:Integer,S,D<:AbstractDict{I,S}} <: AbstractTracer
```

src/tracers.jl
```diff
@@ -133,31 +145,45 @@ function Base.show(io::IO, t::HessianTracer{S}) where {S}
...
+
+HessianTracer{S,I}(::Number) where {S,I} = empty(HessianTracer{S,I})
 HessianTracer(t::HessianTracer) = t
 
+function keys2set(::Type{S}, d::Dict{I}) where {I<:Integer,S<:AbstractSet{<:I}}
+    return S(keys(d))
+end
+function keys2set(::Type{S}, d::Dict{I}) where {I<:Integer,S<:SortedVector{I}}
+    return S(collect(keys(d)); sorted=false)
+end
```
gdalle, line 179: this is another slow bit, right?

src/tracers.jl
gdalle: good one

test/benchmark.jl
```diff
@@ -6,7 +6,7 @@ using NNlib: conv
 
 include("brusselator.jl")
 
-function benchmark_brusselator(N::Integer, method=:tracer)
+function benchmark_brusselator(N::Integer, method=:tracer_bitset)
```
gdalle, line 9: we could directly give the set type instead of a `method` symbol, at least outside of symbolics

src/sortedvector.jl
adrhill: Yeah, I see the issue now. 👍

src/sortedvector.jl
adrhill: I removed it because it never got called. Where would you need it?

src/tracers.jl
adrhill: It's a good argument for using `SortedDict` with `SortedVector`.

test/benchmark.jl
adrhill: We should just make use of the ADTypes interface here. CC #46.

gdalle: Opened an issue to keep track #48

src/sortedvector.jl
gdalle: Not sure

adrhill:
Merging now and leaving conversations unresolved for a follow-up PR.

gdalle:
closed the remaining conversations because I recorded the concerns in https://github.com/adrhill/SparseConnectivityTracer.jl/issues/40

PR #46, opened 2024-05-05
Split tests to improve modularity and debloat runtests.jl

gdalle:
This is strictly copy-paste and moving stuff around


test/brusselator.jl
```diff
@@ -1,20 +1,29 @@
...
+    A = 1.0
+    B = 1.0
+    alpha = 1.0
+    xyd = fill(1.0, N)
+    dx = 1.0
+    p = (A, B, alpha, xyd, dx, N)
+
+    u = rand(dims...)
+    du = similar(u)
+    f!(du, u) = brusselator_2d_loop(du, u, p, nothing)
```
adrhill, line 19: Since you're already doing this, we could make the Brusselator call one function in `brusselator_definition.jl` that takes two arguments: `N` and an ADTypes sparsity-pattern detector.

This way, all Brusselator code would be abstracted away in one file and could be called through one function.

adrhill (approved):

adrhill:
I don't see a downside in updating the "real-world tests" to make use of the ADTypes interface.

adrhill:
Feel free to merge if you just want to quickly get this done.

test/brusselator.jl
gdalle: I'll do this in a future PR, good idea

PR #53, opened 2024-05-07
Refactor tracer types

### LINKED ISSUE #40: HessianTracer type improvements

adrhill:
Closes #40 by updating tracers to parametric signatures
* `ConnectivityTracer{I,S}`
* `JacobianTracer{I,S}`
* `HessianTracer{I,S,D}`

Where `I` is the index type, `S` is a (pseudo-)set type and `D` is an `AbstractDict{I,S}`.

### DIFF
--- a/src/SparseConnectivityTracer.jl
+++ b/src/SparseConnectivityTracer.jl
@@ -13,6 +13,7 @@ include("overload_hessian.jl")
 include("pattern.jl")
 include("adtypes.jl")
 
+include("settypes/base_sets.jl")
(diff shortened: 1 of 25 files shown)


adrhill:
Merging for now, will be changed in #57.

Issue #57, opened 2024-05-09
Refactor following theory


gdalle:
- [x] Refactor global tracers
- [ ] #56 

--- 

Dumping the code from our meeting
```julia
struct GradientTracer end
struct LocalGradientTracer end

⊕(::AbstractSparseVector, ::AbstractSparseVector)

firstder_firstarg_nonzero(::typeof(max)) = true
firstder_firstarg_nonzero(::typeof(max), α, β) = β > α

for op in all_the_ops
    @eval begin
        function $op(t1::T, t2::T) where {T<:GlobalGradientTracer}
            gα = gradient_sparsity(t1)
            gβ = gradient_sparsity(t2)
            a = firstder_firstarg_nonzero(♣)
            b = firstder_secondarg_nonzero(♣)
            if a && b
                return gα ⊕ gβ
            elseif a
                return gα
            elseif b
                return bβ
            else
                return nothing # empty
            end
        end

        function $op(t1::T, t2::T) where {T<:GlobalHessianTracer}
            gα = gradient_sparsity(t1)
            gβ = gradient_sparsity(t2)
            Hα = hessian_sparsity(t1)
            Hβ = hessian_sparsity(t2)
            a = firstder_firstarg_nonzero(op)
            b = firstder_secondarg_nonzero(op)
            c = seconder_firstard_nonzero(op)
            d = nothing #
            e = nothing #
            # more stuff
        end

        function $op(t1::T, t2::T) where {T<:LocalGradientTracer}
            α = primal(t1)
            β = primal(t2)
            gα = gradient_sparsity(t1)
            gβ = gradient_sparsity(t2)
            a = firstder_firstarg_nonzero(♣, α, β)
            b = firstder_secondarg_nonzero(♣, α, β)
            if a && b
                return gα ⊕ gβ
            elseif a
                return gα
            elseif b
                return bβ
            else
                return nothing # empty
            end
        end
    end
end
```

adrhill:
One key insight from the ongoing refactor in #59 is that we previously used dicts of sets to represent **both** `hessian_sparsity` (key => non-empty set) and `gradient_sparsity` (key => empty set).

Introducing a `GlobalHessianTracer` with two parametric types
```Julia
struct GlobalHessianTracer{G,H} <: AbstractHessianTracer
    grad::G     # sparse binary vector representation of non-zero entries in the gradient
    hessian::H  # sparse binary matrix representation of non-zero entries in the Hessian
end
```
[...]

PR #59, opened 2024-05-10
Refactor code to match theory

### LINKED ISSUE #56: Add local tracers
Currently, all tracers return conservative global sparsity patterns.
Local sparsity can be computed by propagating a primal value as well.

adrhill:
Will close #56, #57.


adrhill:
Up to [`8ed3f4d`](https://github.com/adrhill/SparseConnectivityTracer.jl/pull/59/commits/8ed3f4d2b8c0a22e48b81f00c08dde8579ead418), this just refactors types. Running intermediate benchmarks due to BenchmarkCI not supporting ARM CPUs.

adrhill:
This is now a working refactor of the global tracers according to the theory we developed.

I think this is a good point for an in-depth review before adding the local tracers @gdalle.

For now, the code only supports **internally** representing Hessians as `Set{Tuple{I, I}` where `I<:Integer`.

src/tracers.jl
```diff
@@ -12,147 +12,142 @@ The provided index set type `S` has to satisfy the following conditions:
...
+# Elementwise OR #
+#================#
+
+## We use ∨ to represent the elementwise OR
+# TODO REVIEW: \vee looks too much like the character 'v' in code
+# Gradient representations
+∨(a::G, b::G) where {G<:AbstractSet} = a ∪ b
+∨(a::G, b::G) where {G<:DuplicateVector} = G(vcat(a.data, b.data))
+∨(a::G, b::G) where {G<:SortedVector} = a ∪ b
+∨(a::G, b::G) where {G<:RecursiveSet} = a ∪ b
```
gdalle, line 32: This can go in other files. Basically the central `tracker.jl` file should read as if it had been written for `AbstractVector{Bool}` or `AbstractMatrix{Bool}`, give or take

src/tracers.jl
```diff
@@ -12,147 +12,142 @@ The provided index set type `S` has to satisfy the following conditions:
...
+sparse_vector(T, index) = T([index])
+sparse_vector(::Type{T}, index) where {T<:DuplicateVector} = T(index)
+sparse_vector(::Type{T}, index) where {T<:SortedVector} = T(index)
+
+#================#
+# Elementwise OR #
+#================#
+
+## We use ∨ to represent the elementwise OR
+# TODO REVIEW: \vee looks too much like the character 'v' in code
```
gdalle, line 27: You're right. Let's spell out the operators then:

- `elementwise_or`
- `outer_product_or`
- `symmetric_outer_product_or` (not essential)

src/tracers.jl
gdalle: In which case:
```julia
elementwise_or(a::AbstractVector{Bool}, b::AbstractVector{Bool}) = a .| b
outer_product_or(a::AbstractVector{Bool}, b::AbstractVector{Bool}) = a .| transpose(b)
```

gdalle: Here, the naked theory happens. Save the magic for the set implementations

src/tracers.jl
```diff
@@ -12,147 +12,142 @@ The provided index set type `S` has to satisfy the following conditions:
...
+# Gradient representations
+∨(a::G, b::G) where {G<:AbstractSet} = a ∪ b
+∨(a::G, b::G) where {G<:DuplicateVector} = G(vcat(a.data, b.data))
+∨(a::G, b::G) where {G<:SortedVector} = a ∪ b
+∨(a::G, b::G) where {G<:RecursiveSet} = a ∪ b
+
+# Hessian representations
+# TODO REVIEW: for now, we only support representing Hessians as a set of tuples
+∨(a::H, b::H) where {I<:Integer,H<:AbstractSet{Tuple{I,I}}} = a ∪ b
```
gdalle, line 36: Again, this can go in another file

src/tracers.jl
```diff
@@ -12,147 +12,142 @@ The provided index set type `S` has to satisfy the following conditions:
...
+## We use ∨ to represent the elementwise OR
+# TODO REVIEW: \vee looks too much like the character 'v' in code
+# Gradient representations
+∨(a::G, b::G) where {G<:AbstractSet} = a ∪ b
+∨(a::G, b::G) where {G<:DuplicateVector} = G(vcat(a.data, b.data))
+∨(a::G, b::G) where {G<:SortedVector} = a ∪ b
+∨(a::G, b::G) where {G<:RecursiveSet} = a ∪ b
+
+# Hessian representations
+# TODO REVIEW: for now, we only support representing Hessians as a set of tuples
```
gdalle, line 35: We can support `AbstractMatrix{Bool}` for free

src/tracers.jl
gdalle: `PairSet` or sth

src/tracers.jl
```diff
@@ -12,147 +12,142 @@ The provided index set type `S` has to satisfy the following conditions:
...
+∨(a::H, b::H) where {I<:Integer,H<:AbstractSet{Tuple{I,I}}} = a ∪ b
+
+## Outer product on gradients
+# Compute `out ∨ (𝟙[∇a] ∨ 𝟙[∇b]ᵀ)` in out
+function outer_product_or!(
+    out::H, a::G, b::G
```
gdalle, line 41: why in-place?

src/operators.jl
```diff
@@ -11,6 +11,8 @@
 ##=================================#
 # Operators for functions f: ℝ → ℝ #
 #==================================#
+function is_firstder_zero_global end
+function is_seconder_zero_global end
```
gdalle, line 15: maybe a shorter option
```julia
der1_nonzero
der2_nonzero
der1_arg1_nonzero
der1_arg2_nonzero
der2_arg1_nonzero
der2_arg2_nonzero
der2_arg12_nonzero
```
We could even drop the `nonzero` everywhere since those are internal traits anyway

src/operators.jl
```diff
@@ -41,6 +43,11 @@ ops_1_to_1_s = (
     # other
     :inv,
 )
+for op in ops_1_to_1_s
+    T = typeof(eval(op))
+    SparseConnectivityTracer.is_firstder_zero_global(::T) = false
+    SparseConnectivityTracer.is_seconder_zero_global(::T) = false
+end
```
gdalle, line 50: vibe check passed :heavy_check_mark:

src/overload_hessian.jl
```diff
@@ -1,128 +1,99 @@
...
+# TODO REVIEW: this keeps popping up so often that think it's worth an abstraction.  
+# It was previously called "promote_order".
+# The overly descriptive name is TEMPORARY.
+function tracer_from_hessian_OR_outer_product_OR_of_gradient(
+    t::T
+) where {T<:GlobalHessianTracer}
+    hessian_out = deepcopy(t.hessian)
+    hessian_out = outer_product_or!(hessian_out, t.gradient, t.gradient)
+    return T(t.gradient, hessian_out)
+end
```
gdalle, line 10: For me this should just be

```julia
elementwise_or(t.hessian, outer_product_or(t.gradient, t.gradient)) 
```

No need for in-place stuff

gdalle: And thus it doesn't need a specific name

src/operators.jl
gdalle: alternative

```julia
has_deri_argj
```

src/overload_hessian.jl
```diff
@@ -1,128 +1,99 @@
...
-        b = promote_order(b)
-        return additive_merge(a, b)
+# ## 2-to-1
+for fn in ops_2_to_1
+    @eval function Base.$fn(a::T, b::T) where {G,H,T<:GlobalHessianTracer{G,H}}
+        gradient_out, hessian_out = if !is_firstder_arg1_zero_global($fn)
+            if !is_firstder_arg2_zero_global($fn)
+                (a.gradient ∨ b.gradient, a.hessian ∨ b.hessian)
+            else
+                (a.gradient, deepcopy(a.hessian))
```
gdalle, line 34: why deepcopy here?

src/tracers.jl
gdalle: on the other hand this makes a lot of code more verbose, and further from the paper

syntax highlighting treats `\vee` differently from `vee` in Julia, so maybe we're good?

in which case you would use `\vee^T` for the outer product maybe

```julia
∨  # vee
v  # v
```

src/overload_hessian.jl
```diff
@@ -1,128 +1,99 @@
...
+            end
+        else # is_firstder_arg1_zero_global
+            if !is_firstder_arg2_zero_global($fn)
+                (b.gradient, deepcopy(b.hessian))
+            else
+                (empty_sparse_vector(G), empty_sparse_matrix(H))
```
gdalle, line 40: is there a need to specify `empty_sparse_vector` when `G` is already an abstract sparse vector?

src/overload_hessian.jl
```diff
@@ -1,128 +1,99 @@
...
+            end
+        end
```
gdalle, line 42: is there a reason to have a first batch of `if / else` and then do in-place modifications? I feel like this makes things more obscure

gdalle (commented):
Vibe check completed, I have left lots of comments :) Thank you for this huge undertaking!

As a side note, the reference tests for printing make this PR more noisy than it has to be. Maybe comparing `string(whatever)` would be enough

benchmark/benchmarks.jl
```diff
@@ -28,13 +28,13 @@ end
...
 
 ## Define Benchmark suite
-suite = BenchmarkGroup()
+SUITE = BenchmarkGroup()
 
 for S in SET_TYPES
     setname = string(S)
     method = TracerSparsityDetector(S)
 
-    J_global = suite["Jacobian"]["Global"]
```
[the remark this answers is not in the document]
adrhill: Yes, I don't have a meaningful benchmark for Hessians.

I am not sure adding the tiny functions from our test-set would make for a useful and informative benchmark.

src/SparseConnectivityTracer.jl
```diff
@@ -13,14 +17,9 @@ include("overload_hessian.jl")
...
-include("settypes/base_sets.jl")
-include("settypes/duplicatevector.jl")
-include("settypes/recursiveset.jl")
-include("settypes/sortedvector.jl")
-
 export ConnectivityTracer, connectivity_pattern
-export JacobianTracer, jacobian_pattern
-export HessianTracer, hessian_pattern
+export GlobalGradientTracer, jacobian_pattern
+export GlobalHessianTracer, hessian_pattern
```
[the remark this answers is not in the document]
adrhill: I plan on adding `jacobian_pattern_local`. Does the ADTypes interface allow to distinguish between local and global patterns yet?

src/tracers.jl
```diff
@@ -12,147 +12,142 @@ The provided index set type `S` has to satisfy the following conditions:
 Subtypes of `AbstractSet{<:Integer}` are a natural choice, like `BitSet` or `Set{UInt64}`.
```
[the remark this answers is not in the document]
adrhill: Yeah, I'm not sure we are ever going to use anything but Sets.

I will update the comment to differentiate between the mathematical concept of a sparse vector and the implementation detail of types like `AbstractSet` or `AbstractVector` in Julia.

src/tracers.jl
```diff
@@ -12,147 +12,142 @@ The provided index set type `S` has to satisfy the following conditions:
...
 """
 
+empty_sparse_vector(T) = T()
+empty_sparse_matrix(T) = T()
+
+sparse_vector(T, index) = T([index])
+sparse_vector(::Type{T}, index) where {T<:DuplicateVector} = T(index)
+sparse_vector(::Type{T}, index) where {T<:SortedVector} = T(index)
```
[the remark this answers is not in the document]
adrhill: I have an alternative proposal:
We should subtype `DuplicateVector{T}`, `SortedVector{T}` and `RecursiveSet{T}` as `AbstractSet{T}`.

This way, tracer code only needs to be written for generic `AbstractSet` and doesn't need to touch `src/settypes`. Vice versa, tracer logic wouldn't leak into `src/settypes`.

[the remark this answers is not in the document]
adrhill: This is very much related to the [comment below](https://github.com/adrhill/SparseConnectivityTracer.jl/pull/59#discussion_r1597667604): 
Ideally, either everything should be an `AbstractSet{<:Integer}` or an `AbstractMatrix{<:Bool}`.

src/tracers.jl
adrhill: Ok, we need to make a call here. 
The cleanest abstraction would be to make everything either a `AbstractSet{<:Integer}` or an `AbstractVector{<:Bool}`.

Based on this decision, we get either `a ∪ b` or  `a .| b` everywhere. Either one single method
```Julia
∨(a::G, b::G) where {G<:AbstractSet{<:Integer}} = a ∪ b
```
or one single method
```Julia
∨(a::G, b::G) where {G<:AbstractVector{Bool}} = a .| b
```
So far, we have only been using index-sets to represent the *mathematical* concept of a sparse vector. I therefore think `AbstractSet{<:Integer}` is the better abstraction. 

Otherwise, we would have to introduce a wrapper type that turns the index-sets we use into `AbstractVector{Bool}`. And element-wise `|` might be very inefficient on an `AbstractVector` which is actually an `AbstractSet` in disguise.

src/tracers.jl
adrhill: Good news: there is also `\bigvee`, which looks a tad more distinguishable:
```julia
⋁ # bigvee
∨ # vee
v # v
V # V
```
What happened to the idea of using `\oplus` ⊕ for the element-wise OR/max?

src/tracers.jl
adrhill: See discussion above.

src/tracers.jl
adrhill: I suggest the following:
1. I replace the inplace mutations with code matching the theory 1-to-1
2. We add meaningful Hessian benchmarks here or in a follow-up PR 
3. We improve performance in another follow-up PR

src/overload_hessian.jl
adrhill: Will do, see [here](https://github.com/adrhill/SparseConnectivityTracer.jl/pull/59#discussion_r1597668250).

src/overload_hessian.jl
adrhill: [Will remove inplace updates](https://github.com/adrhill/SparseConnectivityTracer.jl/pull/59#discussion_r1597668250)

src/overload_hessian.jl
adrhill: Some two-argument functions have {zero first- and second-order derivatives} on both input, e.g. `div`:
```
julia> div(2.1, 1.3)
1.0

julia> div(2.1, 1.33)
1.0

julia> div(2.12, 1.33)
1.0
```
The code above essentially creates an empty tracer. For comparison:
```Julia
function empty(::Type{GlobalHessianTracer{G,H}}) where {G,H}
    return GlobalHessianTracer{G,H}(empty_sparse_vector(G), empty_sparse_matrix(H))
end
```

src/overload_hessian.jl
adrhill: Let's look at the following equation (screenshot since GitHub fails to render it):
image

The "first batch" computes the union of $\nabla^2\alpha$ and $\nabla^2\beta$.
The second block then updates that allocated Hessian with the four outer-ORs that follow.

src/operators.jl
adrhill: Some behind the scenes on the name selection:

I enjoy the Julia convention of boolean functions starting with `is*`. It makes it easy to guess the return type and `if/else` statement read like english sentences.

Since we need to distinguish between locally and globally zero, we might just as well add it the name for the sake of clarity instead of just distinguishing the two through their amount of arguments. 
And since `nonzero_global` doesn't make sense in the context of sparsity, I opted for `zero_global`.

I have no strong feelings about anything between `is` and `zero_global(ly)`.

benchmark/benchmarks.jl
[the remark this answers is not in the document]
gdalle: I'll try to come up with an example for the Hessian that is similar to "multiplication by sparse matrix" for the Jacobian

src/SparseConnectivityTracer.jl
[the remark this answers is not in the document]
gdalle: not yet, so let's keep them, you're right

src/tracers.jl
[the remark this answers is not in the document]
gdalle: you're right, let's make everything an `AbstractSet{<:Integer}`

src/tracers.jl
gdalle: see above, you're right, let's go `AbstractSet{<:Integer}` all the way (even if we don't implement all the features of an abstract set

gdalle: and `AbstractSet{Tuple{I,I}}` for Hessians?

src/tracers.jl
gdalle: If we switch to abstract sets everywhere, vee becomes a union so the worry disappears.

As for the oplus, since we're only using booleans, a max-plus analogy was overkill. Max for booleans is just OR

src/tracers.jl
gdalle: I'm 100% in favor of that

src/operators.jl
gdalle: no strong feelings here either, but I would like to replace `first` and `second` by `1` and `2` nonetheless, if only because they are the same length

src/overload_hessian.jl
gdalle: what I mean is that `empty(G)` is just as specific as `empty_sparse_vector(G)`

src/overload_hessian.jl
gdalle: I wonder if there is a way to cleverly represent the four outer-ORs together

but as you said above, let's stick to the theory first and optimize later

benchmark/benchmarks.jl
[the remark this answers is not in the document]
adrhill: Good idea, we could actually reuse exactly that for the Hessian.

src/overload_hessian.jl
adrhill: Good point, will simplify

src/overload_hessian.jl
adrhill: As a reference for later:

Since for sets, the outer-ORs correspond to nested loops creating index-tuples `(i, j)`, it might be worth merging $\mathbb{1}[\nabla\alpha] \vee \mathbb{1}[\nabla\alpha]^T$ with $\mathbb{1}[\nabla\alpha] \vee \mathbb{1}[\nabla\beta]^T$ 
to $\mathbb{1}[\nabla\alpha] \vee (\mathbb{1}[\nabla\alpha] \vee \mathbb{1}[\nabla\beta])^T$ when possible.

adrhill: And now that I'm thinking about it, terms 5 and 6 should always correspond to
$(\mathbb{1}[\nabla\alpha] \vee \mathbb{1}[\nabla\beta]) \vee (\mathbb{1}[\nabla\alpha] \vee \mathbb{1}[\nabla\beta])^T$

adrhill: Alternatively, the matrices in term 5 and 6 can be created simultaneously, since they create index tuples `(i , j)` and `(j, i)` respectively.

benchmark/benchmarks.jl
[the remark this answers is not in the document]
gdalle: not exactly I feel, because the following function is not very interesting in terms of tracers mixing (the second order ones only emerge at the very last step):

```julia
function f(x; l, p)
    n = length(x)
    y = copy(x)
    for _ in 1:l
        A = sprand(n, n, p)
        y = A * x
    end
    return dot(x, y)
end
```
but maybe we can adapt it

src/overload_hessian.jl
gdalle: The terms 5 and 6 must definitely be handled together, since they are associated to the same partial derivative

I just wrote them separately in the math because of formatting

benchmark/benchmarks.jl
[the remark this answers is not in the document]
adrhill: As mentioned in another comment, let's add this in a follow-up.

adrhill:
Looks like I broke `DuplicateVector`'s performance @gdalle.

src/operators.jl
```diff
@@ -41,6 +43,11 @@ ops_1_to_1_s = (
     # other
     :inv,
 )
+for op in ops_1_to_1_s
+    T = typeof(eval(op))
+    SparseConnectivityTracer.is_firstder_zero_global(::T) = false
```
gdalle, line 48: why prefix with the package name every time?

src/overload_hessian.jl
```diff
@@ -1,128 +1,85 @@
...
 
 ## 2-to-1
-# Including second-order only
-for fn in ops_2_to_1_ssc
-    @eval function Base.$fn(a::HessianTracer, b::HessianTracer)
-        a = promote_order(a)
-        b = promote_order(b)
-        return distributive_merge(a, b)
+for fn in ops_2_to_1
+    @eval function Base.$fn(a::T, b::T) where {G,H,T<:GlobalHessianTracer{G,H}}
```
gdalle, line 18: this is perfect

src/tracers.jl
```diff
@@ -1,194 +1,184 @@
...
 
-- it is an iterable with `<:Integer` element type
-- it implements `union`
+#================#
+# Elementwise OR #
+#================#
 
-Subtypes of `AbstractSet{<:Integer}` are a natural choice, like `BitSet` or `Set{UInt64}`.
-"""
+×(a::G, b::G) where {G<:AbstractSet} = Set((i, j) for i in a, j in b)
```
gdalle, line 16: what about
```julia
Set(Base.Iterators.product(a, b))
```

src/tracers.jl
```diff
@@ -1,194 +1,184 @@
...
+```
 """
-struct ConnectivityTracer{I<:Integer,S} <: AbstractTracer
-    inputs::S # indices of connected, enumerated inputs
-end
-function ConnectivityTracer(inputs::S) where {S}
-    I = eltype(S)
-    return ConnectivityTracer{I,S}(inputs)
+struct ConnectivityTracer{C<:AbstractSet{<:Integer}} <: AbstractTracer
+    inputs::C # sparse binary vector representing non-zero indices of connected, enumerated inputs
```
gdalle, line 40: you can put this between quotes above the field and use DocStringExtensions.jl to automatically add it to the struct docstring

src/tracers.jl
```diff
@@ -1,194 +1,184 @@
...
 end
 
-function Base.show(io::IO, t::ConnectivityTracer{I,S}) where {I,S}
+function Base.show(io::IO, t::ConnectivityTracer)
     return Base.show_delim_array(
-        io, convert.(Int, inputs(t)), "ConnectivityTracer{$I,$S}(", ',', ')', true
+        io, convert.(Int, sort!(collect(t.inputs))), "$(typeof(t))(", ',', ')', true
```
gdalle, line 45: won't this modify the object in-place when `collect` does nothing?

gdalle (approved):
Awesome! Let's get this merged, figure out local sparsity and then split hairs on performance

src/tracers.jl
adrhill: Is that more performant / type stable?

Otherwise I find the verbose option a bit more readable. Afaik, it calls `Base.Iterators.product` as well.

src/tracers.jl
adrhill: Nice catch. It probably does.

src/operators.jl
adrhill: To define methods in the scope of the module. Is there another way? Removing the module name, I get the error 
```
ERROR: LoadError: UndefVarError: `T` not defined
Stacktrace:
 [1] top-level scope
   @ ~/Developer/SparseConnectivityTracer.jl/src/operators.jl:46
```

gdalle: I spontaneously would have tried something like
```julia
@eval is_firstder_yero_global(::$(typeof(op)))
```
and manipulated the actual operations instead of symbols. But that's very minor and can easily be fixed so don't let it stop you

src/tracers.jl
gdalle: then let's keep it, I have no specific argument for either

PR #65, opened 2024-05-15
Add local tracers

### LINKED ISSUE #56: Add local tracers

adrhill:
Closes #56.

* Add a `Dual` wrapper type keeping track of primal computations
  * rename `GlobalGradientTracer` back to `GradientTracer`
  * rename `GlobalHessianTracer` back to `HessianTracer`

Still WIP.

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -208,7 +208,7 @@ for op in ops_2_to_1_fsc
 end
 
 # gradient of x/y: [1/y -x/y²]
-SparseConnectivityTracer.is_firstder_arg2_zero_local(::typeof(Base.:/)) = iszero(x)
+SparseConnectivityTracer.is_firstder_arg2_zero_local(::typeof(Base.:/), x, y) = iszero(x)
 
 # ops_2_to_1_fsz: 
 # ∂f/∂x    != 0
@@ -245,9 +245,9 @@ for op in ops_2_to_1_ffc
 end
 
 # gradient of x*y: [y x]
-SparseConnectivityTracer.is_firstder_arg1_zero_local(::typeof(Base.:*)) = iszero(y)
-SparseConnectivityTracer.is_firstder_arg2_zero_local(::typeof(Base.:*)) = iszero(x)
-SparseConnectivityTracer.is_crossder_zero_local(::typeof(Base.:*))      = iszero(x) || iszero(y)
+SparseConnectivityTracer.is_firstder_arg1_zero_local(::typeof(Base.:*), x, y) = iszero(y)
+SparseConnectivityTracer.is_firstder_arg2_zero_local(::typeof(Base.:*), x, y) = iszero(x)
+SparseConnectivityTracer.is_crossder_zero_local(::typeof(Base.:*), x, y)      = iszero(x) || iszero(y)
 
 # ops_2_to_1_ffz: 
 # ∂f/∂x    != 0
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -269,7 +269,6 @@ for op in ops_2_to_1_ffz
 end
 
 is_firstder_arg2_zero_local(::typeof(mod), x, y) = ifelse(y > 0, y > x, x > y)
-is_firstder_arg2_zero_local(::typeof(rem), x, y) = ifelse(y > 0, y > x, x > y)
 
 is_firstder_arg1_zero_local(::typeof(max), x, y) = x < y
 is_firstder_arg2_zero_local(::typeof(max), x, y) = y < x
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -82,27 +82,10 @@ for op in ops_1_to_1_z
     SparseConnectivityTracer.is_seconder_zero_global(::T) = true
 end
 
-# Functions returning constant output
-# that only depends on the input type.
-# For the purpose of operator overloading,
-# these are kept separate from ops_1_to_1_z.
-ops_1_to_1_const = (
-    :zero, :one,
-    :eps, 
-    :typemin, :typemax,
-    :floatmin, :floatmax, :maxintfloat, 
-)
-for op in ops_1_to_1_const
-    T = typeof(eval(op))
-    SparseConnectivityTracer.is_firstder_zero_global(::T) = true
-    SparseConnectivityTracer.is_seconder_zero_global(::T) = true
-end
-
 ops_1_to_1 = union(
     ops_1_to_1_s, 
     ops_1_to_1_f, 
     ops_1_to_1_z,
-    ops_1_to_1_const,
 )
 
 ##==================================#
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -101,8 +101,8 @@ function is_crossder_zero_global end
 # Fallbacks for local derivatives:
 is_firstder_arg1_zero_local(f::F, x, y) where {F} = is_firstder_arg1_zero_global(f)
 is_seconder_arg1_zero_local(f::F, x, y) where {F} = is_seconder_arg1_zero_global(f)
-is_firstder_arg2_zero_local(f::F, x, y) where {F} = is_firstder_arg1_zero_global(f)
-is_seconder_arg2_zero_local(f::F, x, y) where {F} = is_seconder_arg1_zero_global(f)
+is_firstder_arg2_zero_local(f::F, x, y) where {F} = is_firstder_arg2_zero_global(f)
+is_seconder_arg2_zero_local(f::F, x, y) where {F} = is_seconder_arg2_zero_global(f)
 is_crossder_zero_local(f::F, x, y) where {F}      = is_crossder_zero_global(f)
 
 # ops_2_to_1_ssc: 
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -105,7 +105,6 @@ ops_1_to_1 = union(
     ops_1_to_1_const,
 )
 
-
 ##==================================#
 # Operators for functions f: ℝ² → ℝ #
 #===================================#
@@ -148,14 +147,14 @@ end
 # ∂²f/∂y²  != 0
 # ∂²f/∂x∂y == 0
 ops_2_to_1_ssz = ()
-for op in ops_2_to_1_ssz
-    T = typeof(eval(op))
-    SparseConnectivityTracer.is_seconder_arg1_zero_global(::T) = false
-    SparseConnectivityTracer.is_firstder_arg1_zero_global(::T) = false
-    SparseConnectivityTracer.is_firstder_arg2_zero_global(::T) = false
-    SparseConnectivityTracer.is_seconder_arg2_zero_global(::T) = false
-    SparseConnectivityTracer.is_crossder_zero_global(::T)      = true
-end
+# for op in ops_2_to_1_ssz
+#     T = typeof(eval(op))
+#     SparseConnectivityTracer.is_seconder_arg1_zero_global(::T) = false
+#     SparseConnectivityTracer.is_firstder_arg1_zero_global(::T) = false
+#     SparseConnectivityTracer.is_firstder_arg2_zero_global(::T) = false
+#     SparseConnectivityTracer.is_seconder_arg2_zero_global(::T) = false
+#     SparseConnectivityTracer.is_crossder_zero_global(::T)      = true
+# end
 
 # ops_2_to_1_sfc: 
 # ∂f/∂x    != 0
@@ -164,14 +163,14 @@ end
 # ∂²f/∂y²  == 0
 # ∂²f/∂x∂y != 0
 ops_2_to_1_sfc = ()
-for op in ops_2_to_1_sfc
-    T = typeof(eval(op))
-    SparseConnectivityTracer.is_firstder_arg1_zero_global(::T) = false
-    SparseConnectivityTracer.is_seconder_arg1_zero_global(::T) = false
-    SparseConnectivityTracer.is_firstder_arg2_zero_global(::T) = false
-    SparseConnectivityTracer.is_seconder_arg2_zero_global(::T) = true
-    SparseConnectivityTracer.is_crossder_zero_global(::T)      = false
-end
+# for op in ops_2_to_1_sfc
+#     T = typeof(eval(op))
+#     SparseConnectivityTracer.is_firstder_arg1_zero_global(::T) = false
+#     SparseConnectivityTracer.is_seconder_arg1_zero_global(::T) = false
+#     SparseConnectivityTracer.is_firstder_arg2_zero_global(::T) = false
+#     SparseConnectivityTracer.is_seconder_arg2_zero_global(::T) = true
+#     SparseConnectivityTracer.is_crossder_zero_global(::T)      = false
+# end
 
 # ops_2_to_1_sfz: 
 # ∂f/∂x    != 0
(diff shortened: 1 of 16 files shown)


src/overload_connectivity.jl
```diff
@@ -1,3 +1,5 @@
+# TODO: support Duals
```
gdalle, line 1: Actually for connectivity they are not needed, since by definition we consider it to be value-independent?

src/overload_gradient.jl
```diff
@@ -1,67 +1,144 @@
...
+    end
+    @eval function Base.$fn(t::D) where {P,T<:GradientTracer,D<:Dual{P,T}}
+        x = primal(t)
+        out = Base.$fn(x)
+        if is_firstder_zero_local($fn, x)
+            return Dual(out, empty(T))
+        else
+            return Dual(out, tracer(t))
+        end
+    end
```
gdalle, line 18: looks good

src/overload_gradient.jl
```diff
@@ -1,67 +1,144 @@
...
+            else # ∂f∂y ≠ 0 
+                return ty
+            end
+        else # ∂f∂x ≠ 0 
+            if ∂f∂y0
+                return tx
+            else # ∂f∂y ≠ 0 
+                return T(gradient(tx) ∪ gradient(ty))
+            end
+        end
```
gdalle, line 38: how about this (schematically) to avoid the nesting?
```julia
left = ∂f∂x0 ? empty(T) : tx
right = ∂f∂y0 ? empty(T) : ty
return left ∪ right
```

src/overload_hessian.jl
```diff
@@ -1,85 +1,206 @@
...
+        if !is_seconder_arg1_zero_local($fn, x, y)
+            union!(hess, gradient(a) × gradient(a))
+        end
+        if !is_seconder_arg2_zero_local($fn, x, y)
+            union!(hess, gradient(b) × gradient(b))
+        end
+        if !is_crossder_zero_local($fn, x, y)
+            union!(hess, (gradient(a) × gradient(b)) ∪ (gradient(b) × gradient(a)))
+        end
+        return Dual(out, T(grad, hess))
```
gdalle, line 77: Everything in this function is identical to the one before, except the last line.
Maybe treatment of ordinary tracers and duals could be unified for the tracing part, and only at the end do you decide what you return?

gdalle: I know there is a difference between `der_zero_local(op, x, y)` and `der_zero_global(op)`, but maybe `primal(xt)` could return `nothing` on a simple `GradientTracer` or something

src/operators.jl
```diff
@@ -249,21 +268,30 @@ for op in ops_2_to_1_ffz
...
 end
 
+is_firstder_arg2_zero_local(::typeof(mod), x, y) = ifelse(y > 0, y > x, x > y)
+is_firstder_arg2_zero_local(::typeof(rem), x, y) = ifelse(y > 0, y > x, x > y)
+
+is_firstder_arg1_zero_local(::typeof(max), x, y) = x < y
+is_firstder_arg2_zero_local(::typeof(max), x, y) = y < x
+
+is_firstder_arg1_zero_local(::typeof(min), x, y) = x > y
+is_firstder_arg2_zero_local(::typeof(min), x, y) = y > x
```
gdalle, line 278: I would use wide inequalities

src/operators.jl
```diff
@@ -249,21 +268,30 @@ for op in ops_2_to_1_ffz
     SparseConnectivityTracer.is_crossder_zero_global(::T)      = true
 end
 
+is_firstder_arg2_zero_local(::typeof(mod), x, y) = ifelse(y > 0, y > x, x > y)
+is_firstder_arg2_zero_local(::typeof(rem), x, y) = ifelse(y > 0, y > x, x > y)
```
gdalle, line 272: I find it weird that these two lines are the same

src/operators.jl
```diff
@@ -196,21 +207,24 @@ for op in ops_2_to_1_fsc
     SparseConnectivityTracer.is_crossder_zero_global(::T)      = false
 end
 
+# gradient of x/y: [1/y -x/y²]
+SparseConnectivityTracer.is_firstder_arg2_zero_local(::typeof(Base.:/), x, y) = iszero(x)
```
gdalle, line 211: it can be just `::typeof(/)`

gdalle (commented):
Haven't read everything but I left some comments to reduce code duplication if possible! Good work :)

src/operators.jl
adrhill: What do you mean?

src/operators.jl
adrhill: I'll take a closer look at `rem` on negative numbers.

src/overload_gradient.jl
adrhill: I'll have a look at which type of control flow statement works best for constant propagation and dead code elimination.

src/overload_connectivity.jl
adrhill: It's necessary for control flow and comparisons. 
Could be useful for [data-flow analysis](https://en.wikipedia.org/wiki/Data-flow_analysis), but isn't really relevant to our goals.

src/operators.jl
gdalle: `<=`

adrhill: It turns out this is correct to support `max(x, y)` where $x = y$:

```Julia-repl
julia> @test local_jacobian_pattern(x -> max(x[1], x[2]), [1.0, 2.0], G) ≈ [0 1;]
Test Passed

julia> @test local_jacobian_pattern(x -> max(x[1], x[2]), [2.0, 1.0], G) ≈ [1 0;]
Test Passed

julia> @test local_jacobian_pattern(x -> max(x[1], x[2]), [1.0, 1.0], G) ≈ [1 1;]
Test Passed
```
```Julia-repl
julia> @test local_jacobian_pattern(x -> min(x[1], x[2]), [1.0, 2.0], G) ≈ [1 0;]
Test Passed

julia> @test local_jacobian_pattern(x -> min(x[1], x[2]), [2.0, 1.0], G) ≈ [0 1;]
Test Passed

julia> @test local_jacobian_pattern(x -> min(x[1], x[2]), [1.0, 1.0], G) ≈ [1 1;]
Test Passed
```

src/operators.jl
adrhill: `:/` is a bit easier to find and jump to than `/`, so I'll leave this.

src/operators.jl
adrhill: Removed `rem` for now.

src/overload_gradient.jl
gdalle: My generic take on this is that any kind of compiler optimization is peanuts compared to the cost of the operations we are doing, which is merging sets. So I would bed on code clarity first and foremost

src/operators.jl
gdalle: right, because the functions are `iszero`

src/overload_connectivity.jl
adrhill: Duals on `ConnectivityTracer` have been added.

adrhill:
#68 is fixed on Jacobians. Hessians and Connectivity still need testing.

src/overload_gradient.jl
```diff
@@ -1,67 +1,149 @@
...
-)
-    @eval Base.$fn(a::T, b::T) where {T<:GlobalGradientTracer} = T(a.grad ∪ b.grad)
-    @eval Base.$fn(t::GlobalGradientTracer, ::Number) = t
-    @eval Base.$fn(::Number, t::GlobalGradientTracer) = t
+## 2-to-1
+function gradient_tracer_2_to_1(
+    tx::T,
+    ty::T,
+    is_firstder_arg1_zero_or_number::Bool,
+    is_firstder_arg2_zero_or_number::Bool,
```
gdalle, line 27: why `or_number`?

src/overload_gradient.jl
```diff
@@ -1,67 +1,149 @@
...
-end
-for fn in ops_2_to_1_zzz
-    @eval Base.$fn(::T, ::T) where {T<:GlobalGradientTracer}      = empty(T)
-    @eval Base.$fn(::T, ::Number) where {T<:GlobalGradientTracer} = empty(T)
-    @eval Base.$fn(::Number, ::T) where {T<:GlobalGradientTracer} = empty(T)
+function gradient_tracer_2_to_1_one_tracer(
+    t::T, is_firstder_zero::Bool
+) where {T<:GradientTracer}
+    # NOTE: this is identical to gradient_tracer_1_to_1 due to ignored second argument having empty set
+    # TODO: remove once gdalle agrees
```
gdalle, line 48: agreed

src/overload_gradient.jl
```diff
@@ -1,67 +1,149 @@
-for fn in union(ops_1_to_1_s, ops_1_to_1_f)
-    @eval Base.$fn(t::GlobalGradientTracer) = t
+## 1-to-1
```
gdalle, line 1: ok

src/overload_gradient.jl
```diff
@@ -1,67 +1,149 @@
...
-    ops_2_to_1_sfz,
-    ops_2_to_1_fsc,
-    ops_2_to_1_fsz,
-    ops_2_to_1_ffc,
-    ops_2_to_1_ffz,
-)
-    @eval Base.$fn(a::T, b::T) where {T<:GlobalGradientTracer} = T(a.grad ∪ b.grad)
-    @eval Base.$fn(t::GlobalGradientTracer, ::Number) = t
-    @eval Base.$fn(::Number, t::GlobalGradientTracer) = t
+## 2-to-1
```
gdalle, line 22: ok

src/overload_gradient.jl
```diff
@@ -1,67 +1,149 @@
...
+        t_out = gradient_tracer_2_to_1_one_tracer(
+            tracer(dy), is_firstder_arg2_zero_local($fn, x, y)
+        )
+        return Dual(p_out, t_out)
+    end
 end
 
-for fn in union(ops_1_to_2_zs, ops_1_to_2_zf)
-    @eval Base.$fn(t::T) where {T<:GlobalGradientTracer} = (empty(T), t)
+## 1-to-2
```
gdalle, line 100: ok

src/overload_gradient.jl
```diff
@@ -1,67 +1,149 @@
...
+function Base.round(
+    d::D, mode::RoundingMode; kwargs...
+) where {P,T<:GradientTracer,D<:Dual{P,T}}
+    return Dual(round(primal(d), mode; kwargs...), empty(T))
+end
 
-## Random numbers
-rand(::AbstractRNG, ::SamplerType{T}) where {T<:GlobalGradientTracer} = empty(T)
+## Random numbers 
+# TODO: support random numbers on Duals
```
gdalle, line 148: what? how? why?

src/overload_hessian.jl
```diff
@@ -1,85 +1,193 @@
...
+        union!(hess, (gradient(a) × gradient(b)) ∪ (gradient(b) × gradient(a)))
+    end
+    return T(grad, hess)
+end
+
+function hessian_tracer_2_to_1_one_tracer(
+    t::T, is_firstder_zero::Bool, is_seconder_zero::Bool
+) where {T<:HessianTracer}
+    # NOTE: this is identical to hessian_tracer_1_to_1 due to ignored second argument having empty set
+    # TODO: remove once gdalle agrees
```
gdalle, line 68: agreed

src/overload_hessian.jl
```diff
@@ -1,85 +1,193 @@
...
-            end
+function hessian_tracer_1_to_1(
+    t::T, is_firstder_zero::Bool, is_seconder_zero::Bool
+) where {T<:HessianTracer}
+    if is_seconder_zero
+        if is_firstder_zero
+            return empty(T)
         else
-            return T(t.grad, t.hess ∪ (t.grad × t.grad))
+            return t
```
gdalle, line 9: I wonder if we shouldn't call the gradient tracer utilities in places like here

src/overload_hessian.jl
```diff
@@ -1,85 +1,193 @@
...
         end
+    else
+        return T(gradient(t), hessian(t) ∪ (gradient(t) × gradient(t)))
```
gdalle, line 12: and here

src/overload_hessian.jl
gdalle: because the non-zeroness of the first derivative can also play a role

src/overload_hessian.jl
```diff
@@ -1,85 +1,193 @@
...
+function hessian_tracer_2_to_1(
+    a::T,
+    b::T,
+    is_firstder_arg1_zero::Bool,
+    is_seconder_arg1_zero::Bool,
+    is_firstder_arg2_zero::Bool,
+    is_seconder_arg2_zero::Bool,
+    is_crossder_zero::Bool,
+) where {G,H,T<:HessianTracer{G,H}}
+    grad = empty(G)
```
gdalle, line 42: here again, should we also trace the gradient with the utilities defined in `overloads_gradient.jl`?

src/overload_hessian.jl
```diff
@@ -1,85 +1,193 @@
 ## 1-to-1
```
gdalle, line 1: not completely ok

src/overload_hessian.jl
```diff
@@ -1,85 +1,193 @@
...
-        if is_seconder_arg1_zero_global($fn)
-            if is_firstder_arg1_zero_global($fn)
-                return empty(T)
-            else
-                return t
-            end
-        else
-            return T(t.grad, t.hess ∪ (t.grad × t.grad))
-        end
+## 2-to-1
```
gdalle, line 32: not completely ok

src/overload_hessian.jl
```diff
@@ -1,85 +1,193 @@
...
+            is_firstder_arg2_zero_local($fn, x, y),
+            is_seconder_arg2_zero_local($fn, x, y),
+        )
+        return Dual(p_out, t_out)
     end
 end
-function Base.:^(t::T, ::Irrational{:ℯ}) where {T<:GlobalHessianTracer}
-    return T(t.grad, t.hess ∪ (t.grad × t.grad))
+
+## 1-to-2
```
gdalle, line 141: ok

gdalle (commented):
I have reviewed the overloads, not the conversion yet. My main doubt is whether we handle gradient tracing properly when we do hessian tracing

src/overload_gradient.jl
adrhill: Old name carried over from a draft, will fix.

src/overload_gradient.jl
adrhill: For global tracers, this was trivial zu support: just return an empty tracer.
For Duals, I agree with dropping it.

src/overload_hessian.jl
adrhill: You're saying we should replace
```Julia
if is_seconder_zero
    if is_firstder_zero
        return empty(T)
    else
        return t
    end
else
    return T(gradient(t), hessian(t) ∪ (gradient(t) × gradient(t)))
end
```

by 
```Julia
if is_seconder_zero
    if is_firstder_zero
        return empty(T)
    else
        return t
    end
else
    if is_firstder_zero
        return T(empty(G), gradient(t) × gradient(t))
    else
        return T(gradient(t), hessian(t) ∪ (gradient(t) × gradient(t)))
    end
end
```
right?

src/overload_hessian.jl
adrhill: Either answered [here](https://github.com/adrhill/SparseConnectivityTracer.jl/pull/65#discussion_r1604009803) or I don't understand your point.

src/overload_hessian.jl
adrhill: Done.

adrhill: `overloads_gradient.jl` only contains definitions on `GradientTracer` and I would have to put unions everywhere save on 2 lines of code duplication.

src/overload_hessian.jl
gdalle: I'm saying Hessian tracing is a strict superset of gradient tracing, so everything we do for gradient tracing we need to re-do here too.
For code clarity I think it might help to split the construction of the `gradient` and `hessian` parts of the output tracer

src/overload_hessian.jl
gdalle: Then let's duplicate. What's important is that we make sure the exact same operations are performed as in the gradient tracing case

gdalle:
> Documenter.jl truly implemented their similar method specifically to drive me insane...

How did you fix it all the previous times?

gdalle:
as I live and breathe

gdalle:
passing tests

src/overload_hessian.jl
adrhill: > For code clarity I think it might help to split the construction of the gradient and hessian parts of the output tracer

Skipping this for now. We can always revisit it later.

PR #71, opened 2024-05-18
Replace operator symbols with operator objects

gdalle:
Changes:

- Replace the lists of symbols with lists of actual objects
- (Not important) Group tests for `SortedVector` so that they don't skew the total number of tests

Benefits:

- No need to extend `SparseConnectivityTracer.iszero_...`, just `iszero_...`
- Hovering on the symbols gives you the docstring, which is very helpful in this file for classification

I suspect this change could also allow us to avoid metaprogramming altogether in the `overloads...` files, but I haven't yet figured out how


adrhill:
Looks good, thanks! :)

PR #75, opened 2024-05-19
Add conj, real, transpose, adjoint on numbers

gdalle:


adrhill:
Is this now tested with #76?

adrhill:
Feel free to merge a minor PR like this one on your own when it's ready.

gdalle:
The operators themselves will be tested but not the `dot` product which was our aim. I'll add something to that effect

test/first_order.jl
```diff
@@ -57,6 +57,9 @@ const FIRST_ORDER_SET_TYPES = (
         @test jacobian_pattern(x -> x^ℯ, 1, G) ≈ [1;;]
         @test jacobian_pattern(x -> ℯ^x, 1, G) ≈ [1;;]
         @test jacobian_pattern(x -> round(x, RoundNearestTiesUp), 1, G) ≈ [0;;]
+
+        # Linear algebra
+        @test jacobian_pattern(x -> dot(x[1:2], x[4:5]), rand(5), G) ≈ [1 1 0 1 1]
```
adrhill, line 62: Can we get a test with local tracers as well?

gdalle: done

PR #76, opened 2024-05-20
Revamp classification tests

gdalle:
This PR replaces the hacky classification tests based on symbols with tests that use the traits defined internally. Essentially, if we declare that a derivative is zero (locally or globally), we check that ForwardDiff agrees. Note that we don't check the other way around, which means we are free to underspecify zeros

### DIFF
--- a/test/classification.jl
+++ b/test/classification.jl
@@ -1,116 +1,76 @@
 using SparseConnectivityTracer:
     ops_1_to_1,
-    ops_1_to_1_s,
-    ops_1_to_1_f,
-    ops_1_to_1_z,
+    is_firstder_zero_global,
+    is_seconder_zero_global,
+    is_firstder_zero_local,
+    is_seconder_zero_local,
     ops_2_to_1,
-    ops_2_to_1_ssc,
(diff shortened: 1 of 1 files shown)


test/classification.jl
```diff
@@ -1,231 +1,150 @@
...
-    try
-        return maximum(classify_1_to_1(op, random_input(op); atol) for _ in 1:trials)
-    catch e
-        @warn "Classification of 1-to-1 operator $op failed" e
-        return error_order
+function correct_classification_1_to_1(op, x; atol)
+    dfdx = derivative(op, x)
+    d²fdx² = second_derivative(op, x)
+    if (is_firstder_zero_global(op) | is_firstder_zero_local(op, x)) &&
+        !isapprox(dfdx, 0; atol)
```
adrhill, line 50: Why don't we test the two separately? Doesn't the `OR` intertwine information?

adrhill: Ok, now I get it, this just quickly returns `false`.

adrhill (approved):
Looks good, thanks! :)

PR #78, opened 2024-05-21
Add influence classification for ConnectivityTracer

gdalle:
- Revamp `ConnectivityTracer` to follow the blueprint of `GradientTracer`, except that it uses `is_influence` instead of `is_firstder`
- Add a new category of operators whose influence is zero, and distinguish them from those who have an influence but zero derivatives
- Handle `big`, `widen`, `zero`, `one`, `oneunit`, `typemin`, `typemax`, `eps`, `float`, `floatmin`, `floatmax`, `maxintfloat` like normal `1_to_1` operators _when they are applied to numbers_ (not when they are applied to types)
- Add special case for when ForwardDiff cannot compute derivatives (we assume those derivatives are zero)

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -66,8 +66,11 @@ ops_1_to_1_f = (
     +, -,
     identity,
     abs, hypot,
-    deg2rad, rad2deg,
-    mod2pi, prevfloat, nextfloat,
+    # angles
+    deg2rad, rad2deg, mod2pi,
+    # floats
+    float, prevfloat, nextfloat,
+    big, widen,
 )
 for op in ops_1_to_1_f
     T = typeof(op)
@@ -98,8 +101,7 @@ end
 ops_1_to_1_i = (
     zero, one, oneunit,
     typemin, typemax, eps,
-    float, floatmin, floatmax, maxintfloat,
-    big, widen,
+    floatmin, floatmax, maxintfloat,
 )
 for op in ops_1_to_1_i
     T = typeof(op)
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -1,7 +1,8 @@
 ## Operator definitions
 
 # We use a system of letters to categorize operators:
-#   z: first- and second-order derivatives (FOD, SOD) are zero
+#   i: independence - no influence at all
+#   z: influence but first- and second-order derivatives (FOD, SOD) are zero
 #   f: FOD ∂f/∂x is non-zero, SOD ∂²f/∂x² is zero 
 #   s: FOD ∂f/∂x is non-zero, SOD ∂²f/∂x² is non-zero
 #   c: Cross-derivative ∂²f/∂x∂y is non-zero
@@ -11,14 +12,17 @@
 ##=================================#
 # Operators for functions f: ℝ → ℝ #
 #==================================#
+function is_influence_zero_global end
 function is_firstder_zero_global end
 function is_seconder_zero_global end
 
 # Fallbacks for local derivatives:
+is_influence_zero_local(f::F, x) where {F} = is_influence_zero_global(f)
 is_firstder_zero_local(f::F, x) where {F} = is_firstder_zero_global(f)
 is_seconder_zero_local(f::F, x) where {F} = is_seconder_zero_global(f)
 
 # ops_1_to_1_s: 
+# x -> f  != 0
 # ∂f/∂x   != 0
 # ∂²f/∂x² != 0
 ops_1_to_1_s = (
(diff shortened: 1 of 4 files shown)


src/operators.jl
```diff
@@ -385,14 +444,18 @@ ops_2_to_1 = union(
...
 function is_seconder_out1_zero_global end
 function is_firstder_out2_zero_global end
 function is_seconder_out2_zero_global end
 
 # Fallbacks for local derivatives:
-is_seconder_out1_zero_local(f::F, x) where {F} = is_seconder_out1_zero_global(f)
+is_influence_out1_zero_local(f::F, x) where {F} = is_influence_out1_zero_global(f)
+is_influence_out2_zero_local(f::F, x) where {F} = is_influence_out2_zero_global(f)
 is_firstder_out1_zero_local(f::F, x) where {F} = is_firstder_out1_zero_global(f)
+is_seconder_out1_zero_local(f::F, x) where {F} = is_seconder_out1_zero_global(f)
```
gdalle, line 458: this was missing

src/operators.jl
```diff
@@ -49,27 +53,34 @@ ops_1_to_1_s = (
...
 ops_1_to_1_f = (
     +, -,
     identity,
     abs, hypot,
-    deg2rad, rad2deg,
-    mod2pi, prevfloat, nextfloat,
+    # angles
+    deg2rad, rad2deg, mod2pi,
+    # floats
+    float, prevfloat, nextfloat,
```
gdalle, line 72: there was a bug here, `float(x)` used to return an empty tracer

src/operators.jl
adrhill: What do you mean?

src/overload_connectivity.jl
```diff
@@ -1,37 +1,125 @@
...
+function connectivity_tracer_2_to_1(
+    tx::T, ty::T, is_influence_arg1_zero::Bool, is_influence_arg2_zero::Bool
+) where {T<:ConnectivityTracer}
+    if is_influence_arg1_zero
+        if is_influence_arg2_zero
+            return empty(T)
+        else
+            return ty
+        end
+    else # ∂f∂x ≠ 0
```
adrhill, line 37: This comment from `GradientTracer` needs to be updated.

src/operators.jl
gdalle: take a look at the source, `is_secondder_out1_zero_local` was never defined

test/classification.jl
```diff
@@ -44,8 +44,16 @@ random_second_input(op) = random_input(op)
...
 
 function correct_classification_1_to_1(op, x; atol)
-    dfdx = derivative(op, x)
-    d²fdx² = second_derivative(op, x)
+    dfdx, d²fdx² = try
+        derivative(op, x), second_derivative(op, x)
+    catch e
+        if e isa MethodError
+            # ForwardDiff can't handle this operator: assume it has zero derivatives
+            0.0, 0.0
```
adrhill, line 52: Why not manually add dispatches on the operator types for zero-derivative operators as we previously did? 
I feel like this would be a bit safer than simply silencing `MethodError`s.

test/classification.jl
```diff
@@ -69,8 +77,16 @@ end;
...
 
 function correct_classification_2_to_1(op, x, y; atol)
-    g = gradient(Base.splat(op), [x, y])
-    H = hessian(Base.splat(op), [x, y])
+    g, H = try
+        gradient(Base.splat(op), [x, y]), hessian(Base.splat(op), [x, y])
+    catch e
+        if e isa MethodError
+            # ForwardDiff can't handle this operator: assume it has zero derivatives
+            zeros(2), zeros(2, 2)
```
adrhill, line 85: Same here.

test/classification.jl
```diff
@@ -115,8 +131,17 @@ function correct_classification_1_to_2(op, x; atol)
...
     end
-    d1 = derivative(op_vec, x)
-    d2 = second_derivative(op_vec, x)
+
+    d1, d2 = try
+        derivative(op_vec, x), second_derivative(op_vec, x)
+    catch e
+        if e isa MethodError
+            # ForwardDiff can't handle this operator: assume it has zero derivatives
+            zeros(2), zeros(2)
```
adrhill, line 140: Same here.

src/operators.jl
adrhill: image

gdalle: oh my bad it wasn't missing, just in a surprising order, I remember now

test/classification.jl
gdalle: done in the last commit

adrhill (approved):
Thanks! :)

PR #82, opened 2024-05-21
Add SpecialFunctions extension

gdalle:


mkitti:
The problematic `@eval` is here:

https://github.com/adrhill/SparseConnectivityTracer.jl/blob/main/src%2Foverload_connectivity.jl#L15

The problem is that you are invoking `overload_connectivity_1_to_1(m, op)` on `__init__` which evals code into the module after it has been precompiled.

The design of SCT us problematic. Instead of defining new methods in the calling module, it defines new methods in SCR. The better way to do this would be to define a macro to extend methods in SCT but define those methods in the calling module.

gdalle:
I put this eval step in the initialization precisely because the error was warning me not to eval stuff during precompilation 🤔 

What do you mean by "define new methods in the calling module"? Should I use the eval of Base/SpecialFunctions?

gdalle:
I finally managed to get it working. Here's a summary of the changes:

**Source**

- Make the `overload_stuff` functions return expressions instead of evaluating them (in `overload_connectivity.jl`, `overload_gradient.jl` and `overload_hessian.jl`. The inspiration is from ForwardDiff: https://github.com/JuliaDiff/ForwardDiff.jl/blob/1907196ba50e3eb337860301ece3ec37d0582257/src/dual.jl#L230-L244
- Create functions `list_operators_i_to_j(::Val{ModuleName})` to list operators in a certain module depending on their arity
- Evaluate the generated expressions all at once in `overload_all.jl`. Again, the inspration is from ForwardDiff: https://github.com/JuliaDiff/ForwardDiff.jl/blob/1907196ba50e3eb337860301ece3ec37d0582257/src/dual.jl#L463-L477

**Extensions**

- Add SpecialFunctions extension with Requires for Julia < 1.9
- Classify the unary and binary operators there
- Add the one-liner `eval(SCT.overload_all(:SpecialFunctions))`

**Tests**

- Test classification for the operators from SpecialFunctions
- Add connectivity, jacobian and hessian tests to check that they can digest tracer numbers

ext/SparseConnectivityTracerSpecialFunctionsExt/SparseConnectivityTracerSpecialFunctionsExt.jl
```diff
@@ -0,0 +1,115 @@
...
+    using SpecialFunctions
+else
+    import ..SparseConnectivityTracer as SCT
+    using ..SpecialFunctions
+end
+
+#=
+Complex functions are ignored.
+Functions with more than 2 arguments are ignored.
+Functions with integer arguments are ignored.
```
adrhill, line 14: Why ignore functions with integer arguments?
While they are non-differentiable, they are still useful for `ConnectivityTracer`.

src/overload_all.jl
```diff
@@ -0,0 +1,30 @@
...
+        end for op in nameof.(list_operators_1_to_2(Val(M)))
+    ]
+    return quote
+        $(exprs_1_to_1...)
+        $(exprs_2_to_1...)
+        $(exprs_1_to_2...)
+    end
+end
+
+eval(overload_all(:Base))
```
adrhill, line 30: Yeah, this combined with `list_operators_*` and the big `quote`s in `overload_*.jl` feels very hacky. 🫤
There are too many layers of metaprogramming with a lot of implicit dependencies in their design that interact across several files.

ext/SparseConnectivityTracerSpecialFunctionsExt/SparseConnectivityTracerSpecialFunctionsExt.jl
gdalle: Because integer arguments require special cases, like the `^` shenanigans we had to deal with in #83. And because I'm not well-versed in SpecialFunctions. So I'd rather have something error than be implemented incorrectly. We can always improve later

src/overload_all.jl
gdalle: It's exactly the way they do it in ForwardDiff:

- The big quotes: https://github.com/JuliaDiff/ForwardDiff.jl/blob/1907196ba50e3eb337860301ece3ec37d0582257/src/dual.jl#L230-L244
- The evaluation: https://github.com/JuliaDiff/ForwardDiff.jl/blob/1907196ba50e3eb337860301ece3ec37d0582257/src/dual.jl#L469-L472
- The listing: https://github.com/JuliaDiff/ForwardDiff.jl/blob/1907196ba50e3eb337860301ece3ec37d0582257/src/dual.jl#L463

gdalle: I spent quite a lot of time trying out various things to make it work with minimal changes, and this seems to work.

gdalle: What's in the quotes is exactly the same as what was directly `eval`-ed earlied (if you ignore the additional module prefixes, which I got from ForwardDiff).
The difference is that we split code generation (creating the expression) from code evaluation (running it through `eval`)

gdalle: The contents of `overload_all` are only designed so that it can be a one-liner in all the package extensions we will no doubt need to add

ext/SparseConnectivityTracerSpecialFunctionsExt/SparseConnectivityTracerSpecialFunctionsExt.jl
adrhill: I agree with the second point, but `^` is related to the exponent having a multitude of different methods depending on argument types, leading to ambiguities if you only specify the concrete type of one of the two arguments.

mkitti:
> * Add the one-liner `eval(SCT.overload_all(:SpecialFunctions))`

Would it be easier to implement this as a macro? A macro is a function that returns an expression and calls `eval`.

gdalle:
> Would it be easier to implement this as a macro? A macro is a function that returns an expression and calls eval.

That's probably a good idea, will try it in a later PR!

Issue #74, opened 2024-05-19
Scalar operators to support


gdalle:
I think we should look at the list on https://docs.julialang.org/en/v1.12-dev/base/math/

gdalle:
According to @ChrisRackauckaus, to transition the SciML ecosystem from Symbolics to SparseConnectivityTracer, we would need to support the equivalent of:

- [DiffRules.jl](https://github.com/JuliaDiff/DiffRules.jl)
- [NaNMath.jl](https://github.com/JuliaMath/NaNMath.jl)
- [DataInterpolations.jl](https://github.com/SciML/DataInterpolations.jl)

adrhill:
DiffRules has a nice list of things we need to support: https://github.com/JuliaDiff/DiffRules.jl/blob/master/src/rules.jl

I guess this means adding a package extension for NaNMath and more SpecialFunctions?
Not sure what DataInterpolations requires, but some of the recently opened issues on `round` were related to it.

adrhill:
With #178, #184 and #187, we are very close to supporting SciML.
Some methods might be missing, but adding them to the existing extensions should be trivial.

PR #85, opened 2024-05-22
Add NNlib extension

adrhill:
Adds NNlib activation functions, as discussed in [this comment](https://github.com/adrhill/SparseConnectivityTracer.jl/issues/56#issuecomment-2112597863) in #56.


ext/SparseConnectivityTracerNNlibExt.jl
```diff
@@ -0,0 +1,97 @@
...
+    @eval SCT.is_influence_zero_global(::$T) = false
+    @eval SCT.is_firstder_zero_global(::$T) = false
+    @eval SCT.is_seconder_zero_global(::$T) = false
+end
+
+SCT.is_seconder_zero_local(::typeof(celu)) = x > 0
+SCT.is_seconder_zero_local(::typeof(elu)) = x > 0
+SCT.is_seconder_zero_local(::typeof(selu)) = x > 0
+
+SCT.is_firstder_zero_local(::typeof(hardswish)) = x < -3
```
gdalle, line 51: the first derivative is never zero?

gdalle (commented):
Missing a few second order tests but otherwise lgtm

As a side note, NNlib activation functions are implemented from Base functions, so if we supported the following properly it would be enough:

- `ifelse`
- `oftype`
- `clamp`
- `muladd`
- `evalpoly`
- `@fastmath` (not sure what it takes to support it)

Of course the local sparsity rules would not kick in, or not as well

adrhill:
> As a side note, NNlib activation functions are implemented from Base functions

To some degree, the argument that everything is made up from Base functions can be applied to any pure Julia code. We're trying to skip some computations.

I was hoping for a visible speed-up in the benchmarks, but they only test global tracing on `Conv` layers with `identity` activation functions.

Benchmarking locally, the difference is very minor on `relu` (2 allocations saved):
```Julia
julia> @benchmark local_jacobian_pattern(LAYER_RELU, $INPUT_FLUX)
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  16.458 μs …  2.417 ms  ┊ GC (min … max): 0.00% … 98.49%
 Time  (median):     18.083 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   18.862 μs ± 38.831 μs  ┊ GC (mean ± σ):  3.96% ±  1.95%

    ▁▄▄▃▁       ▁▂▄▅▇▆█▇▆▅▃▂▁                                 ▂
  ▃▅███████▆▇▇█████████████████▇████████▇▅██▇▇▇▇▆▆▆▂▆▅▅▆▆▄▆▆▆ █
  16.5 μs      Histogram: log(frequency) by time      21.1 μs <

 Memory estimate: 23.91 KiB, allocs estimate: 355.
```
```Julia
julia> @benchmark local_jacobian_pattern(LAYER_RELU, $INPUT_FLUX)
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  16.625 μs …  2.329 ms  ┊ GC (min … max): 0.00% … 98.01%
 Time  (median):     18.000 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   18.891 μs ± 37.138 μs  ┊ GC (mean ± σ):  3.76% ±  1.95%

            ▁▂▄▅▅▇██▇▆▄▃▂▁▁▁▁▁  ▁ ▁▁▁▁                        ▂
  ▃▁▁▁▃▆▅▇▇████████████████████▇████████████▇█▆▇▇▆▇▆▆▇▆▆▇▆▆▆▅ █
  16.6 μs      Histogram: log(frequency) by time      21.4 μs <

 Memory estimate: 23.78 KiB, allocs estimate: 353.
```

For other activation functions that are more involved than `relu`, an increase in performance should be much more noticeable.

adrhill:
> so if we supported the following properly it would be enough

We already support and test on `ifelse`:

https://github.com/adrhill/SparseConnectivityTracer.jl/blob/57def1e7ea26f4412dcb0693463807ed5633e5ca/test/second_order.jl#L190-L197

Adding `clamp`, `muladd` and friends would open the can of worms that are three argument `3_to_1` functions. These have even more cases of differentiability than the already complicated `2_to_1` functions, in case we want to support them in a generic fashion.

If we ever support them, then probably just by computing the most conservative estimate of the Jacobian/Hessian. Or we simply add the minimum required code for overloads on `clamp` and `muladd` specifically.

ext/SparseConnectivityTracerNNlibExt.jl
adrhill: `hardswish(x) = x * hardσ(x)` and 
`hardσ(x) = max(0, min(1, (x + 3) / 6))`, so we get a function that's zero until -3, quadratic until 3 and linear afterwards:

https://www.wolframalpha.com/input?i=x+*+max%280%2C+min%281%2C+%28x+%2B+3%29+%2F+6%29%29

gdalle: right, my bad

gdalle:
I'm not against the extension, I think it's a good idea. Just using the opportunity to highlight that we're missing some stuff from Base, but #74 is perfect to track it. Feel free to merge!

adrhill:
I just realized another upside of adding extensions: it allows non-dual tracers to run through functions that otherwise call `ifelse`, `max` etc.!

gdalle:
They should already be able to flow through `ifelse` and `max`

adrhill:
You're right about `max`, but `ifelse` requires a boolean primal.

PR #92, opened 2024-05-22
Make `AbstractTracer` subtype of `Real`

### LINKED ISSUE #91: Make `AbstractTracer` a subtype of `Real`, not `Number`
Increases compatibility with existing code

adrhill:
Make `AbstractTracer` subtype of `Real` instead of `Number`. Closes #91.

Not sure whether we need to consider this as breaking, since tracer types are considered internals. However, the documentation mentioned tracers subtyping `Number`.


adrhill:
> Not sure whether we need to consider this as breaking, since tracer types are considered internals.

This should be considered breaking because SCT might stop working on some functions it previously worked on.

gdalle:
why did you add `Complex` everywhere?

Issue #90, opened 2024-05-22
Global sparsity detection with control flow


gdalle:
We need to handle both branches of `ifelse` so that global sparsity patterns may be reused at different points `x`

gdalle:
Example where the sparsity pattern depends on whether `x[1] <= 0`:

https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/main/src/ADNLPProblems/AMPGO07.jl

The trouble is that we need to explore both branches to give a global pattern that holds for different values of `x`

adrhill:
We can support `ifelse`, but I don't yet see how we could handle direct comparisons `x <= 0` and hit all branches of the program using just operator overloading.

adrhill:
The real issue is having to decide what comparisons like `tracer > 42` return on non-dual tracers. To support `ifelse`, like I did in #94, the answer has to be a tracer. Otherwise we can't overload and dispatch on `ifelse`. 

However, having comparisons return tracers could lead to all kinds of problems, including code invalidations in Julia Base. And without comparisons on non-dual tracers, control flow is basically useless.

Note that all of this is just to support `ifelse`. Regular if-else-blocks can't be supported via operator overloading.

gdalle:
> The real issue is having to decide what comparisons like tracer > 42 return on non-dual tracers. To support ifelse, like I did in https://github.com/adrhill/SparseConnectivityTracer.jl/pull/94, the answer has to be a tracer. Otherwise we can't overload and dispatch on ifelse.

I actually had a different overloading scheme in mind:

- `<(x::Tracer, y::Tracer) = missing`
- `ifelse(::Missing, x::Tracer, y::Tracer) = union(x, y)`

gdalle:
https://discourse.julialang.org/t/how-bad-is-it-if-x-y-doesnt-return-a-bool/114611

gdalle:
I think your way of returning a tracer would be better, because it also allows the quantity `x < y` to be used as a number, like they do in https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/main/src/ADNLPProblems/AMPGO07.jl

adrhill:
> https://discourse.julialang.org/t/how-bad-is-it-if-x-y-doesnt-return-a-bool/114611

I mirror a lot of the worries that have been expressed in here. 
I really view this package as binary second-order ForwardDiff and not as Symbolics.jl.

[...]

PR #94, opened 2024-05-22
Support `ifelse`

adrhill:
Towards #90.

### DIFF
--- a/src/overload_ifelse.jl
+++ b/src/overload_ifelse.jl
@@ -1,0 +1,28 @@
+const SimpleTracer = Union{ConnectivityTracer,GradientTracer,HessianTracer}
+
+function Base.ifelse(::T, x, y) where {T<:SimpleTracer}
+    size(x) != size(y) &&
+        throw(DimensionMismatch("Outputs sizes of `ifelse` arguments don't match in size."))
+    return output_union(x, y)
+end
+
+## output union on scalar outputs
+output_union(tx::C, ty::C) where {C<:ConnectivityTracer} = C(inputs(tx) ∪ inputs(ty))
+output_union(tx::G, ty::G) where {G<:GradientTracer} = G(gradient(tx) ∪ gradient(ty))
+function output_union(tx::H, ty::H) where {H<:HessianTracer}
+    return H(gradient(tx) ∪ gradient(ty), hessian(tx) ∪ hessian(ty))
+end
+
+output_union(tx::T, y) where {T<:SimpleTracer} = tx
+output_union(x, ty::T) where {T<:SimpleTracer} = ty
+
+## output union on AbstractArray outputs
+function output_union(tx::AbstractArray{T}, ty::AbstractArray{T}) where {T<:SimpleTracer}
+    return output_union.(tx, ty)
+end
+function output_union(tx::AbstractArray{T}, y::AbstractArray) where {T<:SimpleTracer}
+    return tx
+end
+function output_union(x::AbstractArray, ty::AbstractArray{T}) where {T<:SimpleTracer}
+    return ty
+end
--- a/test/test_gradient.jl
+++ b/test/test_gradient.jl
@@ -63,6 +63,17 @@ NNLIB_ACTIVATIONS = union(NNLIB_ACTIVATIONS_S, NNLIB_ACTIVATIONS_F)
         @test jacobian_sparsity(x -> ℯ^x, 1, method) ≈ [1;;]
         @test jacobian_sparsity(x -> round(x, RoundNearestTiesUp), 1, method) ≈ [0;;]
 
+        # ifelse
+        @test jacobian_sparsity(
+            x -> ifelse(x[2], x[1], ifelse(x[1], x[3], x[4])), rand(4), method
+        ) == [1 0 1 1]
+
+        @test jacobian_sparsity(
+            x -> ifelse(x[2], [x[1], x[2]], ifelse(x[1], [x[2], x[3]], [x[3], x[4]])),
+            rand(4),
+            method,
+        ) == [1 1 1 0; 0 1 1 1]
+
         # Linear Algebra
         @test jacobian_sparsity(x -> dot(x[1:2], x[4:5]), rand(5), method) == [1 1 0 1 1]
 
--- a/src/SparseConnectivityTracer.jl
+++ b/src/SparseConnectivityTracer.jl
@@ -24,6 +24,7 @@ include("overload_connectivity.jl")
 include("overload_gradient.jl")
 include("overload_hessian.jl")
 include("overload_dual.jl")
+include("overload_ifelse.jl")
 include("overload_all.jl")
 
 include("pattern.jl")
--- a/test/test_hessian.jl
+++ b/test/test_hessian.jl
@@ -145,6 +145,15 @@ const SECOND_ORDER_SET_TYPES = (
             0 0 0 1 0
             0 1 0 0 1
         ]
+        
+        # ifelse
+        h = hessian_pattern(x -> ifelse(x[1], x[1]^x[2], x[3] * x[4]), rand(5), method)
+        @test h == [
+            1  1  0  0
+            1  1  0  0
+            0  0  0  1
+            0  0  1  0
+        ]
 
         # SpecialFunctions
         @test hessian_sparsity(x -> erf(x[1]), rand(2), method) == [


gdalle:
The test failures you're seeing here are an example of https://github.com/adrhill/SparseConnectivityTracer.jl/pull/95#discussion_r1611575498

adrhill:
> The test failures you're seeing here are an example of comparison method ambiguities [#95 (comment)](https://github.com/adrhill/SparseConnectivityTracer.jl/pull/95#discussion_r1611575498)

I think it's due to me forgetting to remove the previous error-throwing method: https://github.com/adrhill/SparseConnectivityTracer.jl/pull/94/commits/e36f43450e08fa3c869e24832219702f9039a7d8

adrhill:
Very odd and generic error on 1.6. Can `ifelse` not be overloaded here?
```
ERROR: LoadError: LoadError: cannot add methods to a builtin function
Stacktrace:
  [1] top-level scope
    @ ~/work/SparseConnectivityTracer.jl/SparseConnectivityTracer.jl/src/overload_ifelse.jl:3
  [2] include(mod::Module, _path::String)
    @ Base ./Base.jl:384
  [3] include(x::String)
    @ SparseConnectivityTracer ~/work/SparseConnectivityTracer.jl/SparseConnectivityTracer.jl/src/SparseConnectivityTracer.jl:1
  [4] top-level scope
    @ ~/work/SparseConnectivityTracer.jl/SparseConnectivityTracer.jl/src/SparseConnectivityTracer.jl:27
```

gdalle:
No clue, never encountered this

src/overload_ifelse.jl
```diff
@@ -0,0 +1,41 @@
+const SimpleTracer = Union{ConnectivityTracer,GradientTracer,HessianTracer}
+
+function Base.ifelse(::T, x, y) where {T<:SimpleTracer}
+    size(x) != size(y) &&
+        throw(DimensionMismatch("Outputs sizes of `ifelse` arguments don't match in size."))
```
gdalle, line 5: Why is this necessary? They don't have to both be arrays, they could be scalars or god knows what else

src/overload_ifelse.jl
```diff
@@ -0,0 +1,41 @@
...
+
+## output union on scalar outputs
+output_union(tx::C, ty::C) where {C<:ConnectivityTracer} = C(inputs(tx) ∪ inputs(ty))
+output_union(tx::G, ty::G) where {G<:GradientTracer} = G(gradient(tx) ∪ gradient(ty))
+function output_union(tx::H, ty::H) where {H<:HessianTracer}
+    return H(gradient(tx) ∪ gradient(ty), hessian(tx) ∪ hessian(ty))
+end
+
+output_union(tx::T, y) where {T<:SimpleTracer} = tx
+output_union(x, ty::T) where {T<:SimpleTracer} = ty
```
gdalle, line 17: be careful to have something more specific for `(Tracer, Dual)` and `(Dual, Tracer)` otherwise this will take precedence and return incorrect results

src/overload_ifelse.jl
```diff
@@ -0,0 +1,41 @@
...
+    return ty
+end
+
+## Overload comparisons
+# We don't return empty tracers since the resulting Bool could be used as a Number. 
+for fn in (:isequal, :isapprox, :isless, :(==), :(<), :(>), :(<=), :(>=))
+    @eval Base.$fn(tx::C, ty::C) where {C<:ConnectivityTracer} = C(inputs(tx) ∪ inputs(ty))
+    @eval Base.$fn(tx::G, ty::G) where {G<:GradientTracer} = G(gradient(tx) ∪ gradient(ty))
+    @eval function Base.$fn(tx::H, ty::H) where {H<:HessianTracer}
+        return H(gradient(tx) ∪ gradient(ty), hessian(tx) ∪ hessian(ty))
```
gdalle, line 36: we don't need the other terms of hessian tracing (outer products of gradients) here?

src/overload_ifelse.jl
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    @eval Base.$fn(tx::T, y::Real) where {T<:SimpleTracer} = tx
+    @eval Base.$fn(x::Real, ty::T) where {T<:SimpleTracer} = ty
```
gdalle, line 40: isn't this already done above?

src/overload_ifelse.jl
```diff
@@ -0,0 +1,41 @@
...
+    return output_union.(tx, ty)
+end
+function output_union(tx::AbstractArray{T}, y::AbstractArray) where {T<:SimpleTracer}
+    return tx
+end
+function output_union(x::AbstractArray, ty::AbstractArray{T}) where {T<:SimpleTracer}
+    return ty
+end
+
+## Overload comparisons
```
gdalle, line 30: can we add them to the binary operators pipeline, but only for global tracing?

test/test_gradient.jl
```diff
@@ -64,6 +64,25 @@ NNLIB_ACTIVATIONS = union(NNLIB_ACTIVATIONS_S, NNLIB_ACTIVATIONS_F)
...
+            x -> ifelse(x[2], x[1], ifelse(x[1], x[3], x[4])), rand(4), method
+        ) == [1 0 1 1]
+
+        @test jacobian_sparsity(
+            x -> ifelse(x[2], [x[1], x[2]], ifelse(x[1], [x[2], x[3]], [x[3], x[4]])),
+            rand(4),
+            method,
+        ) == [1 1 1 0; 0 1 1 1]
+
+        function f_ampgo07(x)
```
gdalle, line 78: you can use something simpler, no need to show off ;)

test/test_gradient.jl
```diff
@@ -77,7 +96,8 @@ NNLIB_ACTIVATIONS = union(NNLIB_ACTIVATIONS_S, NNLIB_ACTIVATIONS_F)
         end
 
         ## Error handling when applying non-dual tracers to "local" functions with control flow
-        @test_throws MissingPrimalError jacobian_sparsity(
+        # TypeError: non-boolean (SparseConnectivityTracer.GradientTracer{BitSet}) used in boolean context
+        @test_throws TypeError jacobian_sparsity(
```
gdalle, line 100: add this test to connectivity and hessian too

src/overload_ifelse.jl
adrhill: We only support functions with scalar and array outputs and 
```Julia
julia> size(1) == size(42.0)
true
```

adrhill: We could also just let it error on the broadcast of `output_union`.

src/overload_ifelse.jl
adrhill: This doesn't work with `Dual`.

src/overload_ifelse.jl
adrhill: I've thought about this a bit and now suspect it's actually the opposite.
`>` and friends are just regular 2-to-1 operators that are return constant 0 or 1 over their input domain. Their derivatives are therefore zero everywhere.

As a reminder, `hessian_tracer_2_to_1` is defined as
``` 
    grad = empty(G)
    hess = empty(H)
    if !is_firstder_arg1_zero
        grad = union(grad, gradient(a)) # TODO: use union!
        union!(hess, hessian(a))
    end
    if !is_firstder_arg2_zero
        grad = union(grad, gradient(b)) # TODO: use union!
        union!(hess, hessian(b))
    end
    if !is_seconder_arg1_zero
        union!(hess, gradient(a) × gradient(a))
    end
    if !is_seconder_arg2_zero
        union!(hess, gradient(b) × gradient(b))
    end
    if !is_crossder_zero
        union!(hess, (gradient(a) × gradient(b)) ∪ (gradient(b) × gradient(a)))
    end
    return T(grad, hess)
```

The correct implementation therefore is 
```Julia
@eval function Base.$fn(tx::H, ty::H) where {G,H,T<:HessianTracer{G,H}}
    return T(gradient(tx) ∪ gradient(ty), empty(H))
end
```

src/overload_ifelse.jl
adrhill: I don't think so? 

`input`,`gradient` and `hessian` don't have fallbacks on `Real`, since we wouldn't know which set type to return.

src/overload_ifelse.jl
adrhill: We should, see [here](https://github.com/adrhill/SparseConnectivityTracer.jl/pull/94#discussion_r1612311999).

src/overload_ifelse.jl
adrhill: We could let this error on the broadcast, but there are some edge cases where this could go wrong. 

Assume we have an (admittedly type unstable) function `ifelse(cond, expr1, expr2)`. The expression `expr1` computes a scalar `HessianTracer`, `expr2` an array of `HessianTracer`s. Simply broadcasting `expr1` and `expr2` would put the Hessian entries of `expr1` into all outputs of `expr2`.

This is why I added the size check `size(expr1) == size(expr2)`.

src/overload_ifelse.jl
adrhill: Actually not as trivial to do as I thought since the overloads of both global and dual tracers happen in the same function `overload_*_2_to_1(M, op)`.

adrhill: I split the `overload_**(M, op)` functions into separate `overload_**_dual(M, op)` functions. However, I still couldn't add the comparison operators to `operators.jl` due to the `overload_all` mechanism.

test/test_gradient.jl
adrhill: Done.

src/overload_ifelse.jl
gdalle: > We only support functions with scalar and array outputs

That's true for the global function whose Jacobian we analyze. But inside of it there may be functions with arbitrary inputs and outputs:

```julia
struct MyStruct
    a
    b
end

function f(x::Vector)
    s = MyStruct(x[1], x[2])
    return s.a + s.b
end
```

I think the error for a missing broadcast of `size` would be more confusing.

If you want to avoid the situation `ifelse(b, x::T1, y::T2)`, we can define only `ifelse(b, x::T, y::T)`

src/overload_ifelse.jl
gdalle: I meant that in the first function, `y` could be a `Dual`, and in the second `x` could be a `Dual`

src/overload_ifelse.jl
gdalle: I was referring to the lines of code commented here: https://github.com/adrhill/SparseConnectivityTracer.jl/pull/94#discussion_r1612017178

src/overload_ifelse.jl
gdalle: Why not split `overload_all`?

src/overload_ifelse.jl
adrhill: Two different tracer types never appear together.
We either create `ConnectivityTracer`, `GradientTracer`, `HessianTracer` or `Dual`. They never appear together, unless we made a huge mistake somewhere.

src/overload_ifelse.jl
adrhill: Resolving because this has been refactored already. The overload of comparison operators and `output_union` are two different functions.

src/overload_ifelse.jl
adrhill: Can do, I didn't want to break your masterpiece. ;)

adrhill: The issue actually lies in `list_operators`, which also needs to be split.

adrhill: I'm not sure it get's more elegant than what we have now. Did you re-review?

gdalle: Yeah, basically we would need `list_operators_tracer` and `list_operators_dual`?

src/overload_ifelse.jl
adrhill: Good point with the structs and potentially missing broadcasts (isn't it a requirement of the `AbstractArray` interface though?).
However, if `expr1` and `expr2` are structs, we have no way to compute an `output_union` either. 

> If you want to avoid the situation ifelse(b, x::T1, y::T2), we can define only ifelse(b, x::T, y::T)

We can't do this since `ifelse(cond, ::AbstractTracer, ::Float64)` is completely valid and a common occurrence (e.g. `relu`).

Types also aren't a solution for my problem in case `expr1` and `expr2` are both arrays of different sizes but can broadcast.

gdalle: Sorry by broadcast I meant dispatch

gdalle: I'm just saying that the `size` check is not useful, if we have arrays the error will happen when trying to broadcast (not dispatch :sunglasses:) them together, and if we have weird objects it will just obscure the actual method that we need to implement

adrhill: > if we have arrays the error will happen when trying to broadcast (not dispatch 😎) them together

This isn't necessarily true, an `expr1` of size `(2, 3, 4)` will happily broadcast with an `expr2` of size `(2, 3)`.

src/overload_ifelse.jl
adrhill: Yeah basically. But then we would also have to split and duplicate our entire list in `operators.jl`. I think this solution is the simplest with the least amount of code.

test/test_connectivity.jl
```diff
@@ -73,10 +73,16 @@ NNLIB_ACTIVATIONS = union(NNLIB_ACTIVATIONS_S, NNLIB_ACTIVATIONS_F)
             @test connectivity_pattern(f, 1, S) ≈ [1;;]
         end
 
-        # Error handling when applying non-dual tracers to "local" functions with control flow
-        @test_throws MissingPrimalError connectivity_pattern(
+        # ifelse
+        @test connectivity_pattern(
             x -> ifelse(x[2] < x[3], x[1] + x[2], x[3] * x[4]), [1 2 3 4], S
```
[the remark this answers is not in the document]
adrhill: It's in L83.

src/overload_ifelse.jl
gdalle: but then that's the behavior we want

gdalle: for arrays of numbers and for arrays of tracers

test/test_gradient.jl
```diff
@@ -64,6 +64,25 @@ NNLIB_ACTIVATIONS = union(NNLIB_ACTIVATIONS_S, NNLIB_ACTIVATIONS_F)
...
+        ) == [1 0 1 1]
+
+        @test jacobian_sparsity(
+            x -> ifelse(x[2], [x[1], x[2]], ifelse(x[1], [x[2], x[3]], [x[3], x[4]])),
+            rand(4),
+            method,
+        ) == [1 1 1 0; 0 1 1 1]
+
+        function f_ampgo07(x)
+            return (x[1] <= 0) * convert(eltype(x), Inf) +
```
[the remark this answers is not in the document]
adrhill: Added!

test/test_connectivity.jl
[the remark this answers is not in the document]
adrhill: Added!

src/overload_ifelse.jl
adrhill: No, we don't want arrays of tracers of size `(2, 3, 4)` to broadcast with arrays of tracers of size `(2, 3)`.

gdalle: why not?

adrhill: Using your example:
```Julia
julia> using SparseConnectivityTracer

julia> foo(x) = reshape(x[1:12], (2, 3, 2));

julia> bar(x) = reshape(x[13:18], (2, 3, 1));

julia> jacobian_pattern(foo, rand(18))
12×18 SparseArrays.SparseMatrixCSC{Bool, Int64} with 12 stored entries:
⎡⠑⢄⠀⠀⠀⠀⠀⠀⠀⎤
⎢⠀⠀⠑⢄⠀⠀⠀⠀⠀⎥
⎣⠀⠀⠀⠀⠑⢄⠀⠀⠀⎦

julia> jacobian_pattern(bar, rand(18))
6×18 SparseArrays.SparseMatrixCSC{Bool, Int64} with 6 stored entries:
⎡⠀⠀⠀⠀⠀⠀⠑⡀⠀⎤
⎣⠀⠀⠀⠀⠀⠀⠀⠈⢄⎦

julia> jacobian_pattern(x -> ifelse(x[1], foo(x), bar(x)), rand(18))  # <--- with size check
ERROR: DimensionMismatch: Outputs sizes of `ifelse` arguments don't match in size.
Stacktrace:
 [1] ifelse
   @ ~/Developer/SparseConnectivityTracer.jl/src/overload_ifelse.jl:2 [inlined]
 [2] (::var"#5#6")(x::Vector{SparseConnectivityTracer.GradientTracer{BitSet}})
   @ Main ./REPL[6]:1
 [3] trace_function(::Type{SparseConnectivityTracer.GradientTracer{BitSet}}, f::var"#5#6", x::Vector{Float64})
   @ SparseConnectivityTracer ~/Developer/SparseConnectivityTracer.jl/src/pattern.jl:32
 [4] jacobian_pattern(f::Function, x::Vector{Float64}, ::Type{BitSet})
   @ SparseConnectivityTracer ~/Developer/SparseConnectivityTracer.jl/src/pattern.jl:201
 [5] top-level scope
   @ REPL[6]:1

julia> jacobian_pattern(x -> ifelse(x[1], foo(x), bar(x)), rand(18))   # <--- without size check
12×18 SparseArrays.SparseMatrixCSC{Bool, Int64} with 24 stored entries:
⎡⠑⢄⠀⠀⠀⠀⠑⢄⠀⎤
⎢⠀⠀⠑⢄⠀⠀⢄⠀⠑⎥
⎣⠀⠀⠀⠀⠑⢄⠀⠑⢄⎦
```

adrhill: It is not immediately clear to me why this Jacobian pattern with "24 stored entries" for 18 inputs is correct.

gdalle: - The first 12 columns of the Jacobian correspond to `x[1:12]`. These components are used only once in the output
- The last 6 columns of the Jacobian correspond to `x[13:18]`. These components are used twice in the output because they are duplicated by broadcasting over the third dimension.

```julia
julia> using ForwardDiff, SparseArrays

julia> function f(x)
           y = reshape(x[1:12], (2, 3, 2))
           z = reshape(x[13:18], (2, 3, 1))
           return y .+ z
       end
f (generic function with 1 method)

julia> ForwardDiff.jacobian(f, rand(18)) |> sparse
12×18 SparseMatrixCSC{Float64, Int64} with 24 stored entries:
⎡⠑⢄⠀⠀⠀⠀⠑⢄⠀⎤
⎢⠀⠀⠑⢄⠀⠀⢄⠀⠑⎥
⎣⠀⠀⠀⠀⠑⢄⠀⠑⢄⎦
```

gdalle: Basically it's legal in Julia to do this so it should be legal in SCT

gdalle: > It is not immediately clear to me why this Jacobian pattern with "24 stored entries" for 18 inputs is correct.

If the Jacobians could only have as many nonzero entries as there are inputs, life would be boring ;)

gdalle: > Exactly. But why is this mathematically correct? Why do we want this behavior?


It's correct because it is what Julia does. The following functions are equivalent

```julia
julia> function f(x)
           y = reshape(x[1:12], (2, 3, 2))
           z = reshape(x[13:18], (2, 3, 1))
           return y .+ z
       end
f (generic function with 1 method)

julia> function g(x)
           y = reshape(x[1:12], (2, 3, 2))
           z = reshape(x[13:18], (2, 3))
           return y .+ cat(z, z; dims=3)
       end
g (generic function with 1 method)

julia> x = rand(18);

julia> f(x) == g(x)
true
```

> Even from a pure "ConnectivityTracer perspective", bar(x) = reshape(x[13:18], (2, 3, 1)) never even "influenced" the outputs at indices 7 to 12.

This is wrong, as evidenced by the ForwardDiff Jacobian (if there is a nonzero in the Jacobian, there is connectivity.

gdalle: > My take is: if m changes depending on the control flow, we throw an error. This is what the size check does.

Your definition of control flow is a bit extreme. By that token, we should error on the following function since `3` is "control flow":

```julia
h(x) = fill(x[1], 3)
```

gdalle: And if you need one more argument, this kind of implicit broadcasting over singleton dimensions is heavily used in deep learning

adrhill: > It's correct because it is what Julia does. 

Julia doesn't do sparsity detection. The question is about the mathematically correct way to do sparsity detection.

> The following functions are equivalent

I agree with your examples, but they are **not** in the context of `ifelse`.
Julia does **not** broadcast the two branches of an `ifelse` function.

> as evidenced by the ForwardDiff Jacobian (if there is a nonzero in the Jacobian, there is connectivity.

Once again, ForwardDiff does not broadcast the two branches of an `ifelse` function.

adrhill: The code in question is:
```Julia
function Base.ifelse(::AbstractTracer, result_of_some_computation, result_of_other_computation)
    size(result_of_some_computation) != size(result_of_other_computation) &&
        throw(DimensionMismatch("Outputs sizes of `ifelse` arguments don't match in size."))
    return compute_union_of_tracers_in_results(result_of_some_computation, result_of_other_computation)
end

Comment (gdalle) [reply src/overload_ifelse.jl:3]: It's not inherently wrong for `ifelse` to choose between two objects of different sizes. Typically if there's a broadcast afterwards that is compatible with both

Comment (gdalle) [reply src/overload_ifelse.jl]: For a less contrived example:

```julia
julia> g(x) = x .+ ifelse(x[1] > 0, x, 1)
g (generic function with 1 method)

julia> g([3, 4])
2-element Vector{Int64}:
 6
 8

julia> g([-3, 4])
2-element Vector{Int64}:
 -2
  5
```

gdalle: It's not blocking for me so if it's really important for you, go ahead. But you'll be forbidding perfectly legal Julia code like the one above, for which both global and local tracing make perfect sense.

PR #95, opened 2024-05-23
Limit Dual to Real primals, replace all Number with Real

### LINKED ISSUE #91: Make `AbstractTracer` a subtype of `Real`, not `Number`

gdalle:
Fix #91 for good

### DIFF
--- a/src/conversion.jl
+++ b/src/conversion.jl
@@ -2,15 +2,15 @@
 
 ## Type conversions (non-dual)
 for TT in (GradientTracer, ConnectivityTracer, HessianTracer)
-    Base.promote_rule(::Type{T}, ::Type{N}) where {T<:TT,N<:Number} = T
-    Base.promote_rule(::Type{N}, ::Type{T}) where {T<:TT,N<:Number} = T
+    Base.promote_rule(::Type{T}, ::Type{N}) where {T<:TT,N<:Real} = T
+    Base.promote_rule(::Type{N}, ::Type{T}) where {T<:TT,N<:Real} = T
 
     Base.big(::Type{T})   where {T<:TT} = T
     Base.widen(::Type{T}) where {T<:TT} = T
 
-    Base.convert(::Type{T}, x::Number)   where {T<:TT} = empty(T)
+    Base.convert(::Type{T}, x::Real)   where {T<:TT} = empty(T)
     Base.convert(::Type{T}, t::T)        where {T<:TT} = t
-    Base.convert(::Type{<:Number}, t::T) where {T<:TT} = t
+    Base.convert(::Type{<:Real}, t::T) where {T<:TT} = t
 
     ## Constants
     Base.zero(::Type{T})        where {T<:TT} = empty(T)
@@ -42,11 +42,11 @@ function Base.promote_rule(::Type{Dual{P1, T}}, ::Type{Dual{P2, T}}) where {P1,P
     PP = Base.promote_type(P1, P2) # TODO: possible method call error?
     return Dual{PP,T}
 end
-function Base.promote_rule(::Type{Dual{P, T}}, ::Type{N}) where {P,T,N<:Number}
+function Base.promote_rule(::Type{Dual{P, T}}, ::Type{N}) where {P,T,N<:Real}
     PP = Base.promote_type(P, N) # TODO: possible method call error?
     return Dual{PP,T}
 end
-function Base.promote_rule(::Type{N}, ::Type{Dual{P, T}}) where {P,T,N<:Number}
+function Base.promote_rule(::Type{N}, ::Type{Dual{P, T}}) where {P,T,N<:Real}
     PP = Base.promote_type(P, N) # TODO: possible method call error?
     return Dual{PP,T}
 end
@@ -54,9 +54,9 @@ end
 Base.big(::Type{D})   where {P,T,D<:Dual{P,T}} = Dual{big(P),T}
 Base.widen(::Type{D}) where {P,T,D<:Dual{P,T}} = Dual{widen(P),T}
 
-Base.convert(::Type{D}, x::Number) where {P,T,D<:Dual{P,T}}           = Dual(x, empty(T))
+Base.convert(::Type{D}, x::Real) where {P,T,D<:Dual{P,T}}           = Dual(x, empty(T))
 Base.convert(::Type{D}, d::D)      where {P,T,D<:Dual{P,T}}           = d
-Base.convert(::Type{N}, d::D)      where {N<:Number,P,T,D<:Dual{P,T}} = Dual(convert(T, primal(d)), tracer(d))
+Base.convert(::Type{N}, d::D)      where {N<:Real,P,T,D<:Dual{P,T}} = Dual(convert(T, primal(d)), tracer(d))
 
 function Base.convert(::Type{Dual{P1,T}}, d::Dual{P2,T}) where {P1,P2,T} 
     return Dual(convert(P1, primal(d)), tracer(d))
--- a/src/overload_gradient.jl
+++ b/src/overload_gradient.jl
@@ -136,7 +136,7 @@ end
 ## Special cases
 
 ## Exponent (requires extra types)
-for S in (Real, Integer, Rational, Irrational{:ℯ})
+for S in (Integer, Rational, Irrational{:ℯ})
     Base.:^(t::GradientTracer, ::S) = t
     Base.:^(::S, t::GradientTracer) = t
 
--- a/src/overload_gradient.jl
+++ b/src/overload_gradient.jl
@@ -69,12 +69,12 @@ function overload_gradient_2_to_1(M, op)
             return $SCT.Dual(p_out, t_out)
         end
 
-        function $M.$op(tx::$SCT.GradientTracer, ::Number)
+        function $M.$op(tx::$SCT.GradientTracer, ::Real)
             return $SCT.gradient_tracer_1_to_1(
                 tx, $SCT.is_firstder_arg1_zero_global($M.$op)
             )
         end
-        function $M.$op(dx::D, y::Number) where {P,T<:$SCT.GradientTracer,D<:$SCT.Dual{P,T}}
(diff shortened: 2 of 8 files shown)


src/tracers.jl
```diff
@@ -191,12 +191,12 @@ HessianTracer(t::HessianTracer) = t
...
 $(TYPEDEF)
 
-Dual number type keeping track of the results of a primal computation as well as a tracer.
+Dual `Real` number type keeping track of the results of a primal computation as well as a tracer.
 
 ## Fields
 $(TYPEDFIELDS)
 """
-struct Dual{P<:Number,T<:Union{ConnectivityTracer,GradientTracer,HessianTracer}} <:
+struct Dual{P<:Real,T<:Union{ConnectivityTracer,GradientTracer,HessianTracer}} <:
```
adrhill, line 199: This is all we need.

adrhill (commented):
I'm not sure we actually need the other changes besides the one I commented on. Would be interested in discussing this.

src/tracers.jl
gdalle: I thought so too. But then I ran the tests and found that leaving the `::Number` dispatches led to some method ambiguities, typically between
```julia
<=(::Real, ::Real)
<=(::SCT.Dual, ::Number)
```

adrhill:
The original motivation for this PR is that #92 introduced a weird edge case:

> Some weirdness resulting from [#92]: our local tracer type Dual can contain a Complex primal value but is Real.
Maybe the most sane fix is to constrain the primal to Real as well.


This PR however removes a lot of functionality from `ConnectivityTracer` and other global tracers without any obvious gain. For example, the following doesn't work anymore:
```Julia
connectivity_pattern(x -> x^(2im), 1)
```

src/tracers.jl
adrhill: I would even argue that it would be more informative to have `MethodError`s due to type restrictions occur here rather than in `Base.promote_rule`.

gdalle:
However, I note that ForwardDiff doesn't constrain the primal and partial values inside their `Dual` struct, only the `Dual` itself must subtype `Real`:

https://github.com/JuliaDiff/ForwardDiff.jl/blob/1907196ba50e3eb337860301ece3ec37d0582257/src/dual.jl#L14-L21

src/tracers.jl
gdalle: See the discussion below, we shouldn't promise complex number support anyway. Tracers are real numbers, and if people want to go beyond they can do `Complex{Tracer}`

gdalle:
I think however that we should be able to support functions that do $\mathbb{R} \to \mathbb{C} \to \mathbb{R}$ out of the box, so the last commit reintroduces the former complex-valued tests by projecting them to the real axis

adrhill:
> the right way to proceed is to make a Complex{Dual}, not a Dual{Complex}.

This I like. I guess this requires adding some new `Base.promote_rule`?

gdalle:
> This I like. I guess this requires adding some new Base.promote_rule?

The test I added back failed and that's probably the reason, will investigate

src/tracers.jl
```diff
@@ -224,16 +224,16 @@ end
...
     return GradientTracer{G}(sparse_vector(G, index))
 end
-function create_tracer(::Type{ConnectivityTracer{C}}, ::Number, index::Integer) where {C}
+function create_tracer(::Type{ConnectivityTracer{C}}, ::Real, index::Integer) where {C}
     return ConnectivityTracer{C}(sparse_vector(C, index))
 end
-function create_tracer(::Type{HessianTracer{G,H}}, ::Number, index::Integer) where {G,H}
+function create_tracer(::Type{HessianTracer{G,H}}, ::Real, index::Integer) where {G,H}
     return HessianTracer{G,H}(sparse_vector(G, index), empty(H))
 end
```
adrhill, line 239: Needs support for returning a `Complex{T}` when called with `Complex` number.

src/tracers.jl
```diff
@@ -211,7 +211,7 @@ gradient(d::Dual{P,T}) where {P,T<:GradientTracer} = gradient(d.tracer)
 gradient(d::Dual{P,T}) where {P,T<:HessianTracer} = gradient(d.tracer)
 hessian(d::Dual{P,T}) where {P,T<:HessianTracer} = hessian(d.tracer)
 
-function Dual{P,T}(x::Number) where {P<:Number,T<:AbstractTracer}
+function Dual{P,T}(x::Real) where {P<:Real,T<:AbstractTracer}
     return Dual(convert(P, x), empty(T))
 end
```
adrhill, line 216: Needs support for returning a `Complex{T}` when called with `Complex` number.

src/pattern.jl
```diff
@@ -258,7 +258,7 @@ function local_jacobian_pattern(f!, y, x, ::Type{G}=DEFAULT_VECTOR_TYPE) where {
 end
 
 function jacobian_pattern_to_mat(
-    xt::AbstractArray{T}, yt::AbstractArray{<:Number}
+    xt::AbstractArray{T}, yt::AbstractArray{<:Real}
 ) where {T<:GradientTracer}
```
adrhill, line 262: Pattern methods need support for `Complex{T}`.

src/pattern.jl
```diff
@@ -166,7 +166,7 @@ function connectivity_pattern_to_mat(
 end
 
 function connectivity_pattern_to_mat(
-    xt::AbstractArray{D}, yt::AbstractArray{<:Number}
+    xt::AbstractArray{D}, yt::AbstractArray{<:Real}
 ) where {P,T<:ConnectivityTracer,D<:Dual{P,T}}
```
adrhill, line 170: Pattern methods need support for `Complex{T}`.

src/pattern.jl
```diff
@@ -15,7 +15,7 @@ Supports [`ConnectivityTracer`](@ref), [`GradientTracer`](@ref) and [`HessianTra
 """
 trace_input(::Type{T}, x) where {T<:AbstractTracer} = trace_input(T, x, 1)
 
-function trace_input(::Type{T}, x::Number, i::Integer) where {T<:AbstractTracer}
+function trace_input(::Type{T}, x::Real, i::Integer) where {T<:AbstractTracer}
```
adrhill, line 18: Needs support for returning a Complex{T} when called with Complex number.

src/overload_hessian.jl
```diff
@@ -187,7 +187,7 @@ end
 ## Special cases
 
 ## Exponent (requires extra types)
-for S in (Real, Integer, Rational, Complex, Irrational{:ℯ})
```
adrhill, line 190: Can be brought back with `Complex{T}`.

src/overload_hessian.jl
```diff
@@ -104,22 +104,22 @@ function overload_hessian_2_to_1(M, op)
             return $SCT.Dual(p_out, t_out)
         end
 
-        function $M.$op(tx::$SCT.HessianTracer, y::Number)
+        function $M.$op(tx::$SCT.HessianTracer, y::Real)
```
adrhill, line 107: All operators can be brought back with `Complex{T}`.

src/overload_gradient.jl
```diff
@@ -136,7 +136,7 @@ end
 ## Special cases
 
 ## Exponent (requires extra types)
-for S in (Real, Integer, Rational, Complex, Irrational{:ℯ})
```
adrhill, line 139: Can be brought back with Complex{T}.

src/conversion.jl
```diff
@@ -2,15 +2,15 @@
...
+    Base.promote_rule(::Type{N}, ::Type{T}) where {T<:TT,N<:Real} = T
 
     Base.big(::Type{T})   where {T<:TT} = T
     Base.widen(::Type{T}) where {T<:TT} = T
 
-    Base.convert(::Type{T}, x::Number)   where {T<:TT} = empty(T)
+    Base.convert(::Type{T}, x::Real)   where {T<:TT} = empty(T)
     Base.convert(::Type{T}, t::T)        where {T<:TT} = t
-    Base.convert(::Type{<:Number}, t::T) where {T<:TT} = t
+    Base.convert(::Type{<:Real}, t::T) where {T<:TT} = t
```
adrhill, line 13: Support `Complex{T}`.

gdalle:
Here's an argument for only supporting real inputs (nevermind the output type):
- we established in #91 that many functions in the ecosystem are limited to `::AbstractArray{<:Real}`
- a `Dual{Complex} <: Real` doesn't make much sense to me

adrhill:
> Here's an argument for only supporting real inputs (nevermind the output type):
> 
> * we established in [Make `AbstractTracer` a subtype of `Real`, not `Number` #91](https://github.com/adrhill/SparseConnectivityTracer.jl/issues/91) that many functions in the ecosystem are limited to `::AbstractArray{<:Real}`
> * a `Dual{Complex} <: Real` doesn't make much sense to me

I agree with all of this. My suggestion was that a `Complex{Float64}` input could be converted to
* `Complex{T}` for non-dual tracers, `T<:Union{ConnectivityTracer, GradientTracer, HessianTracer}`.
* `Complex{Dual{Float64, T}}` for dual tracers.

adrhill:
This would require additional `trace_input` and `create_tracer` methods for `Complex` numbers.

adrhill:
But all of this is not something that has to be done in this PR (or has to be done in general until the need for it arises).

src/tracers.jl
gdalle: #102

src/pattern.jl
gdalle: #102

src/pattern.jl
gdalle: #102

src/pattern.jl
gdalle: #102

src/overload_hessian.jl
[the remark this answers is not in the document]
gdalle: indeed

src/tracers.jl
gdalle: #102

src/conversion.jl
gdalle: #102

Issue #100, opened 2024-05-23
Weird ambiguity on 1.6 but not on 1.10


gdalle:
From #83 testing (see the [log](https://github.com/adrhill/SparseConnectivityTracer.jl/actions/runs/9205934476/job/25322715961)):

```julia
MethodError: +(::SparseConnectivityTracer.Dual{HessianTracer{BitSet, Set{Tuple{Int64, Int64}}}, HessianTracer{BitSet, Set{Tuple{Int64, Int64}}}}, ::SparseConnectivityTracer.Dual{Float64, HessianTracer{BitSet, Set{Tuple{Int64, Int64}}}}) is ambiguous. Candidates:
    +(x::Real, dy::D) where {P, T<:HessianTracer, D<:SparseConnectivityTracer.Dual{P, T}} in SparseConnectivityTracer at /home/runner/work/SparseConnectivityTracer.jl/SparseConnectivityTracer.jl/src/overload_hessian.jl:132
    +(dx::D, y::Real) where {P, T<:HessianTracer, D<:SparseConnectivityTracer.Dual{P, T}} in SparseConnectivityTracer at /home/runner/work/SparseConnectivityTracer.jl/SparseConnectivityTracer.jl/src/overload_hessian.jl:122
  Possible fix, define
    +(::Union{D, D, D, D}, ::Union{D, D, D, D}) where {P, T<:HessianTracer, D<:SparseConnectivityTracer.Dual{P, T}, P, T<:HessianTracer, D<:SparseConnectivityTracer.Dual{P, T}, T<:HessianTracer, D<:SparseConnectivityTracer.Dual{P, T}, T<:HessianTracer, D<:SparseConnectivityTracer.Dual{P, T}, T<:HessianTracer, D<:SparseConnectivityTracer.Dual{P, T}, D<:SparseConnectivityTracer.Dual{P, T}, D<:SparseConnectivityTracer.Dual{P, T}, T<:HessianTracer, D<:SparseConnectivityTracer.Dual{P, T}}
  Stacktrace:
    [1] add_sum(x::SparseConnectivityTracer.Dual{HessianTracer{BitSet, Set{Tuple{Int64, Int64}}}, HessianTracer{BitSet, Set{Tuple{Int64, Int64}}}}, y::SparseConnectivityTracer.Dual{Float64, HessianTracer{BitSet, Set{Tuple{Int64, Int64}}}})
      @ Base ./reduce.jl:27
[...]

PR #101, opened 2024-05-23
Constrain type of primal value in `Dual`

adrhill:
The `primal` of a `Dual` tracer should never be a tracer itself. 
This will lead to a more informative error in #100.

### DIFF
--- a/src/tracers.jl
+++ b/src/tracers.jl
@@ -200,7 +200,18 @@ struct Dual{P<:Number,T<:Union{ConnectivityTracer,GradientTracer,HessianTracer}}
        AbstractTracer
     primal::P
     tracer::T
+
+    function Dual{P,T}(
+        primal::P, tracer::T
+    ) where {P<:Number,T<:Union{ConnectivityTracer,GradientTracer,HessianTracer}}
+        if P <: AbstractTracer
+            error("Primal value of Dual tracer can't be an AbstractTracer.")
+        end
+        return new{P,T}(primal, tracer)
+    end
 end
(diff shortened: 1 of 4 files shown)


src/tracers.jl
```diff
@@ -200,7 +200,18 @@ struct Dual{P<:Number,T<:Union{ConnectivityTracer,GradientTracer,HessianTracer}}
        AbstractTracer
     primal::P
     tracer::T
+
+    function Dual{P,T}(
+        primal::P, tracer::T
+    ) where {P<:Number,T<:Union{ConnectivityTracer,GradientTracer,HessianTracer}}
+        if P <: AbstractTracer
```
gdalle, line 207: wanna forbid `P <: Dual` too while we're at it?

gdalle: also this needs testing

gdalle (changes requested):

src/tracers.jl
gdalle: to check that an error is thrown, and prevent regressions

adrhill: Writing tests right now.
`P <: Dual` is tested since `Dual <: AbstractTracer`.

adrhill: I used the occasion to update our tests with our usual type abbreviations `C`, `G`, `H` and `S` for sets (instead of the previous `G`, which is taken for `GradientTracer`).

gdalle: good one, I wanted to do that ever since you moved the connectivity tests away from the gradient ones

PR #103, opened 2024-05-24
Allow more generic hessian set types

gdalle:
> This PR is not done yet

**Source**

- Add two variants of `connectivity_tracer_i_to_j`, `gradient_tracer_i_to_j` and `hessian_tracer_i_to_j`: one for `AbstractTracer` and one for `AbstractSet`
- Use `gradient_tracer_i_to_j` inside `hessian_tracer_i_to_j` for separation of concerns
- Allow more generic eltype in the various custom sets, so that they can hold `Tuple{Int,Int}`
- Start implementing clever `union!` and `product`, but we will need the more sophisticated `union_product!` I introduced for Hessians

**Tests**

- Add some tests for cartesian products
- Start adding tests for Hessian tracers with `H != Set{Tuple{Int,Int}}`

PR #110, opened 2024-05-31
Fix `float`, fix constructor tests

adrhill:


test/runtests.jl
```diff
@@ -18,11 +18,12 @@ DocMeta.setdocmeta!(
     recursive=true,
 )
 
-GROUP = get(ENV, "JULIA_SCT_TEST_GROUP", "All")
+GROUP = get(ENV, "JULIA_SCT_TEST_GROUP", "Core")
```
adrhill, line 21: FYI @gdalle, this changes the default local test suite to run `Core` instead of `All`.

test/runtests.jl
```diff
@@ -62,16 +64,18 @@ GROUP = get(ENV, "JULIA_SCT_TEST_GROUP", "All")
...
     end
 
-    @testset verbose = true "Simple examples" begin
-        if GROUP in ("Core", "All")
+    if GROUP in ("Core", "All")
+        @info "Testing simple examples..."
+        @testset verbose = true "Simple examples" begin
             @testset "Tracer Construction" begin
-                include("test_connectivity.jl")
+                include("test_constructors.jl")
```
adrhill, line 78: We probably want this fix to be merged into #103, sorry about that!

PR #112, opened 2024-05-31
Fix failing benchmarks

gdalle:


adrhill:
We should add the erroring Arrow-head structure to the package tests.

gdalle:
> We should add the erroring Arrow-head structure to the package tests.

Not sure what you mean by that but it has rien à voir avec la choucroute comme on dit chez moi ^^

Issue #107, opened 2024-05-27
Shared sets and sets with shared objects


gdalle:
To avoid excessive allocations in Hessian tracing, we can make a new set type holding a reference to a shared set, such that every intermediate scalar quantity shares the same pattern (as opposed to having individual patterns).

This would require:

- Overloading `trace_input` so that each index `x[i]` points to the same shared set
- Handling the empty set intelligently to avoid re-creating shared sets in operations like `similar` and `rand`
- Once #103 is merged, overloading $\times$

Note that such shared objects would also be useful to implement tape-based recursive sets

gdalle:
From the testing perspective, we would need to add functions with dead ends to highlight the differences:

- a shared Hessian pattern will include dead ends (see Walther 2008)
- individual Hessian patterns will ignore them

adrhill:
Added in #135.

Issue #118, opened 2024-06-05
Benchmarking against JuMP


gdalle:
Here are the results of a benchmark between SCT and JuMP sparsity detection, on the suite of OptimizationProblems. The benchmark code is at https://github.com/gdalle/SparsityDetectionComparison

Provided that @amontoison validates the benchmark as fair, we are:

- uniformly faster than JuMP for Jacobians
- often much slower (up to almost 100x) for Hessians, but I would say that's expected and partly fixable

![Jacobian](https://github.com/adrhill/SparseConnectivityTracer.jl/assets/22795598/8fd8dc89-1b4f-47cb-8be4-3a88ad64a29e)
![Hessian](https://github.com/adrhill/SparseConnectivityTracer.jl/assets/22795598/ab008060-df16-428b-afe4-44bfcce8aec2)

Vaibhavdixit02:
>but I would say that's expected and partly fixable

Can you say a bit more about this?

Vaibhavdixit02:
>Damn, Vaibhav is watching the repo like a hawk.

😂 yeah you guys are pretty much helping solve problems that bugged me for a year so I try to keep up (I know my integration attempt might not make it seem like that but I have good reasons for the delay sorry 😭)

> including some that would sacrifice some accuracy for huge speedups.

This would be tremendously useful in some cases where the pattern changes and it's recomputed (so it doesn't need to be very accurate)

gdalle:
That's precisely the approach we're taking in our paper with Adrian. Our gamble is to make sparsity pattern detection so fast that it's even worth it when you do it every time due to pattern changes.

adrhill:
> that's expected and probably fixable. Stay tuned!

Agreed!

As Guillaume already mentioned, we so far focussed on correctness over performance. In our current implementation, each scalar value (each tracer) carries an index set that indicates non-zero derivatives. To obtain accurate results over branching compute graphs, these sets are never mutated in place. Even within this accurate approach, we have very low-hanging to pick, like smarter unions over (partially) empty index sets (#80).

A big performance gain could come from adding methods on the array-level, e.g. to LinearAlgebra (#115). Currently, operations like dense matrix-vector products of tracers dispatch down to scalar operations. Instead, we could be a lot smarter in the way we compute unions of index sets.

[...]

PR #119, opened 2024-06-05
Generalize sparsity pattern representations

### LINKED ISSUE #80: Add `isempty` flag to tracers
Since empty tracers are constructed by functions like `similar`, `zero` & friends, empty tracers occur often enough to make it worthwhile to add a binary `isempty` flag. This can be used to skip unions and immediately return the non-zero tracer.

### LINKED ISSUE #114: Generalize first- and second-order information in `HessianTracer`
Currently, first- and second-order information are kept in separate `AbstractSet` and `AbstractSet{<:Tuple}` types.

The code should be generalized to support "shared" representations of second order information, similar to the dictionary of sets pre #57 and global "tape" with shared objects proposed in #107.

adrhill:
Closes #114.
Closes #80.
___

This PR relaxes tracer field types and adds an additional layer of dispatch to `X_tracer_Y_to_Z` functions by adding `X_tracer_Y_to_Z_inner`, which dispatches on the field type. 

It also adds a generic `isempty` field to tracers, which is handled in `X_tracer_Y_to_Z`, agnostic to the field types.

To use custom tracers, both the native interface and the ADTypes.jl interface now expect a full tracer type to be passed, e.g. `TracerSparsityDetector(gradient_tracer_type, hessian_tracer_type)`.
This is therefore tagged as a breaking release.

This PR has also lead to the discovery of compile time problems, which have been described here: https://github.com/adrhill/SparseConnectivityTracer.jl/issues/120#issuecomment-2168186240

To catch this, some tests have been updated and refactored.
___

<details>
<summary>Outdated PR description</summary>

This PR adds a layer of Abstraction by introducing `AbstractSparsityPattern`s.

Over the last dozens of PRs, we've kept refactoring the Tracer type signatures to accommodate new representations for sparse vectors and sparse matrices, most notably:
* moving away from shared first- and second-order information that were dictionaries of sets (#57)
* introducing custom data types, which had to be forced into the `AbstractSet` interface (#50)
* making second-order information more generic (#103)

The introduction of flexible first- and second-order sparsity pattern abstractions in `AbstractFirstOrderPattern` and `AbstractSecondOrderPattern` should facilitate shared sets (#107) and allow for the reintroduction of second-order `DictIndexset`s in the future.

Recommended reading order for reviews:
1. `indexsets.jl` (could use a better name)
1. `tracers.jl` 
1. updated overload files

</details>


src/indexsets.jl
```diff
@@ -0,0 +1,78 @@
...
+"""
+$(TYPEDEF)
+
+First order sparsity pattern represented by an index set of non-zero values.
+Represented by an `AbstractSet` of integer indices.
+
+## Fields
+$(TYPEDFIELDS)
+"""
+struct SetIndexset{S<:AbstractSet{<:Integer}} <: AbstractFirstOrderPattern
```
adrhill, line 30: Since I anticipate this weird struct name will get a comment:

This is a mathematical index set represented by a Julia `AbstractSet` ($\rightarrow$ `SetIndexset`).
As mentioned in the PR description, I plan on adding an index set represented by a Julia `Dict` ($\rightarrow$ `DictIndexset`) in a follow-up PR.

src/overload_hessian.jl
```diff
@@ -3,15 +3,29 @@
...
+    else
+        pattern = hessian_tracer_1_to_1_pattern(
+            t.pattern, is_firstder_zero, is_secondder_zero
+        )
+        return T(pattern) # return tracer
+    end
 end
 
-function hessian_tracer_1_to_1(
+function hessian_tracer_1_to_1_pattern(
```
adrhill, line 16: New pattern types can be supported by adding `*_tracer_*_to_*_pattern` methods.

adrhill:
#118 made me too curious about #80, so I've added the handful of missing lines in 13582fe.

adrhill:
Looks like `NLPModels` tests failed at the same place after one hour of runtime twice in a row:
```
[ Info: Testing Jacobian sparsity for britgas
Error: The operation was canceled.
```

gdalle:
> after one hour of runtime

https://github.com/adrhill/SparseConnectivityTracer.jl/blob/178681523632cd92a7ea8168d0bc64ef43f4a28d/.github/workflows/CI.yml#L18

src/overload_hessian.jl
gdalle: Do we need the three layers `tracer -> pattern -> set`?

src/tracers.jl
```diff
@@ -1,23 +1,9 @@
-abstract type AbstractTracer <: Real end
-
-sparse_vector(T, index::Integer) = T([index])
+abstract type AbstractTracer{I<:AbstractSparsityPattern} <: Real end
```
gdalle, line 1: call it P goddammit

src/tracers.jl
```diff
@@ -48,38 +34,41 @@ Set{Int64} with 2 elements:
...
-end
-
 # We have to be careful when defining constructors:
 # Generic code expecting "regular" numbers `x` will sometimes convert them 
 # by calling `T(x)` (instead of `convert(T, x)`), where `T` can be `ConnectivityTracer`.
 # When this happens, we create a new empty tracer with no input pattern.
-function ConnectivityTracer{C}(::Real) where {C<:AbstractSet{<:Integer}}
-    return myempty(ConnectivityTracer{C})
+function ConnectivityTracer{P}(::Real) where {P<:AbstractFirstOrderPattern}
+    return myempty(ConnectivityTracer{P})
```
gdalle, line 64: where is `myempty` defined for tracers?

gdalle: ok it's below

src/tracers.jl
```diff
@@ -240,6 +219,10 @@ end
 # Utilities #
 #===========#
 
+myempty(::Type{ConnectivityTracer{P}}) where {P} = ConnectivityTracer{P}(myempty(P), true)
+myempty(::Type{GradientTracer{P}}) where {P} = GradientTracer{P}(myempty(P), true)
+myempty(::Type{HessianTracer{P}}) where {P} = HessianTracer{P}(myempty(P), true)
```
gdalle, line 224: In the test suite, we want to check that this is allocation-free. I am not that far but I assume we have the same overloads for `myempty` on common set and element types so that it points to a constant empty set?

gdalle (changes requested):

src/overload_hessian.jl
adrhill: It allows us to handle `isempty` on the tracer level instead of having to reimplement it for each pattern type.

src/indexsets.jl
```diff
@@ -0,0 +1,78 @@
+abstract type AbstractSparsityPattern end
```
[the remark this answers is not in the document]
adrhill: Not sure we'd gain anything from it. Index sets based on `Int`s aren't the only viable representation of sparsity patterns.

src/indexsets.jl
adrhill: In my mind, a mathematical index set is just one specific way to represent a sparsity pattern.

src/indexsets.jl
```diff
@@ -0,0 +1,78 @@
...
+"""
+$(TYPEDEF)
+
+Second order sparsity pattern represented by index sets of non-zero values.
+Represented by two `AbstractSet`s of integer indices ``i`` and tuples ``(i, j)``.
+
+## Fields
+$(TYPEDFIELDS)
+"""
+struct DualSetIndexset{
```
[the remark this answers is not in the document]
adrhill: Agree on avoiding `Dual*`.

We can do `SetSecondOrderIndexset` and `SetFirstOrderIndexset`, but I'm sure we can come up with something better.

src/indexsets.jl
```diff
@@ -0,0 +1,78 @@
...
+    F<:AbstractSet{<:Integer},S<:AbstractSet{<:Tuple{Integer,Integer}}
```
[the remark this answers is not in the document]
adrhill: I think removing subtypes is fine, since we can just document required methods in in the docstring of the pattern.

src/interface.jl
```diff
@@ -74,8 +74,9 @@ julia> connectivity_pattern(f, x)
  ⋅  ⋅  1
 ```
 """
-function connectivity_pattern(f, x, ::Type{C}=DEFAULT_VECTOR_TYPE) where {C}
-    xt, yt = trace_function(ConnectivityTracer{C}, f, x)
+function connectivity_pattern(f, x, ::Type{C}=DEFAULT_FIRST_ORDER_SET_TYPE) where {C}
```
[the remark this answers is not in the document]
adrhill: Yes, I was trying to keep the diff minimal.

src/tracers.jl
adrhill: Nope, constant empty sets were removed in #57 without losing notable amounts of performance. We can try reintroducing them in a follow-up, now that we have benchmarks?

src/indexsets.jl
gdalle: Sure but just name it something that ends with `Pattern`, otherwise this gets very confusing when you bring it together with _actual_ set types. `IndexSetFirstOrderPattern` is fine

src/indexsets.jl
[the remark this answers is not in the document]
gdalle: As stated above, `IndexSetFirstOrderPattern` and `IndexSetSecondOrderPattern` seem perfect to me

src/indexsets.jl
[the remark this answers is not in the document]
gdalle: It's all the more coherent since we don't fully implement the abstract set API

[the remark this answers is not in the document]
adrhill: Is anything needed besides `myempty`, `union`, `union!`, `product` and `Base.show`? 
Are `iterate` and `collect` mandatory?

src/indexsets.jl
adrhill: In my eyes, renaming an `Indexset` to `Pattern` is just as non-informative as renaming `Float32` to `Number`.

How about `IndexSet` and `SecondOrderIndexSet`?

src/indexsets.jl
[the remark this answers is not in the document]
adrhill: How about `IndexSet` and `SecondOrderIndexSet`?

src/indexsets.jl
adrhill: Keep in mind that you will want to add some sort of `SharedIndexSet` in (#107). This is why I wanted to differentiate this from `IndexSet` by using `SetIndexset`.

adrhill: > How about IndexSet and SecondOrderIndexSet?
Or `SimpleIndexSet`, since using `AbstractSet`s is the most natural way to implement them? But now we don't even restrict to `AbstractSet` anymore...

src/indexsets.jl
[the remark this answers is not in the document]
adrhill: Let's keep the discussion in https://github.com/adrhill/SparseConnectivityTracer.jl/pull/119#discussion_r1628314757.

src/indexsets.jl
adrhill: https://en.wikipedia.org/wiki/Index_set

src/indexsets.jl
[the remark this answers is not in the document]
gdalle: Union of products is nice but not necessary.
For the end step either iterate or collect is needed, iterate is probably better

src/indexsets.jl
adrhill: The pattern can be represented by an https://en.wikipedia.org/wiki/Index_set, which is what we currently do. An alternative would be a mask like a BitMatrix.
Both are very different kinds of sparsity pattern representations.

gdalle: I know!
The first kind should be called IndexSetPattern, the second kind BitMatrixPattern.

gdalle: If you didn't want such long names, you shouldn't have introduced a third layer between tracers and sets

adrhill: Ok, maybe I misunderstood, I thought you were arguing for shorter names.

adrhill: > The first kind should be called IndexSetPattern, the second kind BitMatrixPattern.

The issue with this is that *mathematical index sets* can be implemented using actual (a) sets, (b) dicts, and (c) shared datastructures. We will therefore end up with 

* a) `SetIndexSetPattern`
* b) `DictIndexSetPattern`
* c) `SharedIndexSetPattern`

adrhill: So my suggestion is so call these
* a) `SimpleIndexSet`
* b) `DictIndexSet`
* c) `SharedIndexSet`

The way we don't call `Float32`s `Float32Number`s.

src/interface.jl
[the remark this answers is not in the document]
adrhill: Hmm, changing this opens up the question on whether these are internals or not. Custom set types are currently part of the `*_pattern` docstrings.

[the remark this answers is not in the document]
adrhill: The interface has been updated and `0.6.0-DEV` indicated.

src/indexsets.jl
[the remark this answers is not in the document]
adrhill: Can you check whether the updated docstrings are ok?

src/indexsets.jl
gdalle: My issue is that when people see something like `SimpleIndexSet`, they might expect it to behave like an `AbstractSet`. But here it's something very different, subtyping `AbstractPattern` and _containing_ one or two set objects.
To correct your analogy, it's as if we defined a new type `BitNumber <: AbstractVector{Bool}`. Sure, a number is a vector of bits, but when people see `BitNumber`, they will expect it to behave like a `Number`, not like an `AbstractVector`.
In the same way, when people see `IndexSet`, they will expect it to behave like a `Set`, not like an `AbstractPattern`. Hence the need to add `Pattern` at the end, and it doesn't hurt anyone.

src/interface.jl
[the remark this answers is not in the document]
gdalle: Good, we're gonna break more stuff and that's okay

src/indexsets.jl
```diff
@@ -0,0 +1,141 @@
...
+* `SparseConnectivityTracer.myempty`
+* `SparseConnectivityTracer.seed`
+* `SparseConnectivityTracer.product`
+* `Base.union`
+* `Base.union!`
+* `Base.iterate`
+* `Base.collect`
+* `Base.show`
+
+All of these methods are already implemented for `AbstractSet`s from Julia Base.
```
gdalle, line 65: That is precisely the kind of thing which adds to the confusion. A pattern _is not_ an `AbstractSet`, it is _implemented with_ one or two `AbstractSet`s. So if you want to add such a sentence, it should be something like
```
All of these methods are already implemented for `AbstractSet`s from Julia Base, which the most common pattern subtypes rely on.
```

gdalle (changes requested):

src/indexsets.jl
adrhill: Now I see your point. 👍

src/indexsets.jl
adrhill: > which is used by the most common pattern subtypes.

No, `AbstractSet`s are **exclusively** used in what you'd call `SimpleIndexSetPattern`. They are NOT used in anything else, e.g. what you'd call
* `SharedIndexSetPattern`
* `DictIndexSetPattern`
* `BitMaskPattern`

And even in `SimpleIndexSetPattern`, they are not enforced.

adrhill:
```
All of these methods are already implemented for `AbstractSet`s
from Julia Base, which is used by the most common `SimpleIndexSetPattern` subtypes.
```
Maybe this is what you're suggesting?

adrhill:
A description of our compile time problems (cause & solution) can be found here: https://github.com/adrhill/SparseConnectivityTracer.jl/issues/120#issuecomment-2168186240

adrhill:
Looks like there is a minor decrease in performance in `BitSet` and `SortedVector` and a major increase in performance in `DuplicateVector` and `RecursiveSet`.

Technically speaking, we don't need `@noinline` on `GradientTracer`, since the compile times weren't horrible. I have no strong opinions on this, what do you think @gdalle?

src/adtypes.jl
```diff
@@ -4,7 +4,7 @@
 Singleton struct for integration with the sparsity detection framework of [ADTypes.jl](https://github.com/SciML/ADTypes.jl).
 
 Computes global sparsity patterns over the entire input domain.
-For local sparsity patterns at a specific input point, use [`TracerLocalSparsityDetector`](@ref).
+Gor local sparsity patterns at a specific input point, use [`TracerLocalSparsityDetector`](@ref).
```
gdalle, line 7: For

src/SparseConnectivityTracer.jl
```diff
@@ -15,6 +15,7 @@ include("settypes/duplicatevector.jl")
 include("settypes/recursiveset.jl")
 include("settypes/sortedvector.jl")
 
+# include("sparsity_patterns.jl")
```
gdalle, line 18: remove

src/adtypes.jl
```diff
@@ -31,23 +31,30 @@ julia> ADTypes.hessian_sparsity(f, rand(4), TracerSparsityDetector())
...
-function ADTypes.jacobian_sparsity(f!, y, x, ::TracerSparsityDetector{G,H}) where {G,H}
-    return jacobian_pattern(f!, y, x, G)
+function ADTypes.jacobian_sparsity(f!, y, x, ::TracerSparsityDetector{TG,TH}) where {TG,TH}
+    return jacobian_pattern(f!, y, x, TG)
 end
 
-function ADTypes.hessian_sparsity(f, x, ::TracerSparsityDetector{G,H}) where {G,H}
-    return hessian_pattern(f, x, G, H)
+function ADTypes.hessian_sparsity(f, x, ::TracerSparsityDetector{TG,TH}) where {TG,TH}
+    return hessian_pattern(f, x, TH)
```
gdalle, line 57: I like this separation

src/overload_connectivity.jl
```diff
@@ -1,20 +1,23 @@
...
+    if isemptytracer(t) # TODO: add test
+        return t
+    else
+        i_out = connectivity_tracer_1_to_1_inner(inputs(t), is_influence_zero)
+        return T(i_out) # return tracer
+    end
 end
 
-function connectivity_tracer_1_to_1(
+function connectivity_tracer_1_to_1_inner(
```
gdalle, line 14: that's not an additional layer, just a renaming, right?

src/overload_connectivity.jl
```diff
@@ -43,17 +46,25 @@ end
 
 ## 2-to-1
 
-function connectivity_tracer_2_to_1(
+@noinline function connectivity_tracer_2_to_1(
```
gdalle, line 49: I'd say if the `@noinline`s are not necessary for connectivity and gradient tracing, don't bother?

src/tracers.jl
```diff
@@ -1,30 +1,19 @@
...
-    if isempty(a)
-        return b
-    elseif isempty(b)
-        return a
-    else
-        return union(a, b)
-    end
-end
+@inline myempty(::Type{S}) where {S<:AbstractSet} = S()
+@inline seed(::Type{S}, i::Integer) where {S<:AbstractSet} = S([i])
```
gdalle, line 8: I wonder if we can be smarter than by creating a vector and then a set

src/tracers.jl
```diff
@@ -1,30 +1,19 @@
...
-function clever_union(a::AbstractSet{I}, b::AbstractSet{I}) where {I}
-    if isempty(a)
-        return b
-    elseif isempty(b)
-        return a
-    else
-        return union(a, b)
-    end
-end
+@inline myempty(::Type{S}) where {S<:AbstractSet} = S()
```
gdalle, line 7: this took lots of time in the `britgas` benchmark for some configurations, so it might be worth revisiting constant empty sets at a later day

gdalle (commented):
looks good to me apart from some minor comments

src/overload_connectivity.jl
adrhill: Sure.

src/tracers.jl
adrhill: We have `isempty` flags now, so we can finally use `undef`. I will do this in a follow-up PR today, I didn't want to make benchmarks uninformative by touching too many things that modify the performance.

src/tracers.jl
adrhill: See https://github.com/adrhill/SparseConnectivityTracer.jl/pull/119#discussion_r1642652914.

adrhill: Adding an `EmptyPattern` type was part of the idea of the original PR, but I guess that will be close enough.

src/overload_connectivity.jl
adrhill: This might be a bit confusing to review:
* `connectivity_tracer_1_to_1_inner` is the original layer, previously called `connectivity_tracer_1_to_1`
* `connectivity_tracer_1_to_1` is a new layer that only handles the `isempty` field

adrhill:
I double-checked and the inlining makes no difference on first-order tracers:
```
`GradientTracer` with `@noinline`:
```julia-repl
first jacobian: 0.756304 seconds (729.08 k allocations: 48.768 MiB, 0.66% gc time, 99.98% compilation time)
second jacobian: 0.000062 seconds (2.05 k allocations: 169.844 KiB)
```

without `@noinline`:
```julia-repl
first jacobian: 0.746296 seconds (846.85 k allocations: 54.726 MiB, 0.76% gc time, 99.97% compilation time)
second jacobian: 0.000059 seconds (2.05 k allocations: 169.844 KiB)
```

gdalle:
your call

adrhill:
Since there is no cost in performance, let's just bring it back so we don't have to worry about it in the future!

Issue #108, opened 2024-05-31
StackOverflow when creating `SparseMatrixCSC` with tracers as values


gdalle:
```julia
julia> using ADTypes, SparseConnectivityTracer

julia> using LinearAlgebra, SparseArrays

julia> detector = TracerLocalSparsityDetector()
TracerLocalSparsityDetector{BitSet, Set{Tuple{Int64, Int64}}}()

julia> f(x) = det(spdiagm(x))
f (generic function with 1 method)

julia> g(x) = logdet(spdiagm(x))
g (generic function with 1 method)

julia> ADTypes.hessian_sparsity(f, randn(3), detector)
3×3 SparseMatrixCSC{Bool, Int64} with 6 stored entries:
 ⋅  1  1
 1  ⋅  1
 1  1  ⋅

julia> ADTypes.hessian_sparsity(g, randn(3), detector)
ERROR: StackOverflowError:
Stacktrace:
  [1] Array
    @ ./boot.jl:477 [inlined]
  [2] Array
    @ ./boot.jl:486 [inlined]
  [3] zeros
    @ ./array.jl:636 [inlined]
  [4] zeros
    @ ./array.jl:632 [inlined]
  [5] Dict{Tuple{Int64, Int64}, Nothing}()
    @ Base ./dict.jl:70
  [6] Set
    @ ./set.jl:45 [inlined]
  [7] myempty
    @ ~/Work/GitHub/Julia/SparseConnectivityTracer.jl/src/tracers.jl:5 [inlined]
  [8] myempty
    @ ~/Work/GitHub/Julia/SparseConnectivityTracer.jl/src/tracers.jl:171 [inlined]
  [9] zero
    @ ~/Work/GitHub/Julia/SparseConnectivityTracer.jl/src/conversion.jl:66 [inlined]
 [10] float(A::Vector{SparseConnectivityTracer.Dual{Float64, SparseConnectivityTracer.HessianTracer{BitSet, Set{Tuple{…}}}}})
    @ Base ./float.jl:1120
 [11] float
    @ ~/.julia/juliaup/julia-1.10.3+0.x64.linux.gnu/share/julia/stdlib/v1.10/SparseArrays/src/sparsematrix.jl:983 [inlined]
 [12] lu(A::SparseMatrixCSC{SparseConnectivityTracer.Dual{…}, Int64}; check::Bool) (repeats 18547 times)
[...]

PR #131, opened 2024-06-19
Add array overloads

### LINKED ISSUE #115: Add `AbstractArray{<:AbstractTracer}` methods for common LinearAlgebra functions
As discussed in #108, some functions like `logdet` require control flow, which currently only works with local tracers:
```Julia
julia> using SparseConnectivityTracer, SparseArrays, LinearAlgebra

julia> hessian_pattern(logdet, rand(3, 3))
ERROR: TypeError: non-boolean (SparseConnectivityTracer.HessianTracer{BitSet, Set{Tuple{Int64, Int64}}}) used in boolean context
Stacktrace:
  [1] generic_lufact!(A::Matrix{SparseConnectivityTracer.HessianTracer{BitSet, Set{…}}}, pivot::RowMaximum; check::Bool)
    @ LinearAlgebra /Applications/Julia-1.10.app/Contents/Resources/julia/share/julia/stdlib/v1.10/LinearAlgebra/src/lu.jl:152
...
```

This could be avoided by adding methods like `LinearAlgebra.logdet(M::AbstractMatrix{<:AbstractTracer})` that return conservative estimates of the corresponding sparsity pattern.

adrhill:
Closes #115.
The scope of this PR is somewhat open-ended. What are some must haves for a first array overload PR @gdalle?


gdalle:
Here are the main things I would wanna see:

- [ ] `*`, `mul!` (unlike `+` and `-`, it can be optimized cause each left-side line and each right-side column is reused many times)
- [ ] `dot`
- [ ] `norm`, `opnorm`
- [ ] `\` (I think your idea of using `pinv` and `*` is a good first step)
- [ ] `inv`, `pinv`
- [ ] `det`, `logdet`, `logabsdet`
- [ ] `eigmax`, `eigmin`
- [ ] `exp`, `^`

I think we can disregard special matrix types and factorizations for now.

gdalle:
> This is due to us abusing float to return a tracer.

I've been wondering about that. What happens if we remove it?

adrhill:
It does indeed look like we might not need `float` on our own test suite anymore.

However, external functions could call `float` on a tracer, as seen in the `lu` call above.

gdalle:
> However, external functions could call float on a tracer, as seen in the lu call above.

Indeed let's keep it. My thought process was that it's ethically dubious not to return an `AbstractFloat` when someone calls `float`. But then again we don't return `Bool`s either when someone calls `<` so :shrug:

adrhill:
Somewhat interestingly, #108 is now fixed on global tracers, but not on local tracers:
```Julia-repl
julia> hessian_pattern(x -> logdet(spdiagm(x)), randn(3))
3×3 SparseMatrixCSC{Bool, Int64} with 9 stored entries:
 1  1  1
 1  1  1
 1  1  1

julia> local_hessian_pattern(x -> logdet(spdiagm(x)), randn(3))
ERROR: StackOverflowError:
Stacktrace:
 ...
 [11] float(S::SparseMatrixCSC{SparseConnectivityTracer.Dual{Float64, SparseConnectivityTracer.HessianTracer{…}}, Int64})
    @ SparseArrays ~/.julia/juliaup/julia-1.10.4+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/SparseArrays/src/sparsematrix.jl:983
 [12] lu(A::SparseMatrixCSC{SparseConnectivityTracer.Dual{…}, Int64}; check::Bool) (repeats 28852 times)
    @ SparseArrays.UMFPACK ~/.julia/juliaup/julia-1.10.4+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/SparseArrays/src/solvers/umfpack.jl:395
 [13] logabsdet(A::SparseMatrixCSC{SparseConnectivityTracer.Dual{Float64, SparseConnectivityTracer.HessianTracer{…}}, Int64})
    @ LinearAlgebra ~/.julia/juliaup/julia-1.10.4+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/LinearAlgebra/src/generic.jl:1676
 [14] logdet(A::SparseMatrixCSC{SparseConnectivityTracer.Dual{Float64, SparseConnectivityTracer.HessianTracer{…}}, Int64})
    @ LinearAlgebra ~/.julia/juliaup/julia-1.10.4+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/LinearAlgebra/src/generic.jl:1701
```

adrhill:
> `\` (I think your idea of using `pinv` and `*` is a good first step)

Any concrete feedback on improving this @gdalle? 

I'm planning to make heavy use of FillArrays.jl, e.g.
https://github.com/JuliaArrays/FillArrays.jl/blob/4f8a966e931208a8a5aa56c909cc5e579044e421/src/fillalgebra.jl#L82

gdalle:
> Somewhat interestingly, https://github.com/adrhill/SparseConnectivityTracer.jl/issues/108 is now fixed on global tracers, but not on local tracers:

That's probably because `float(A)` expects to create an `AbstractMatrix{<:AbstractFloat}`. But we don't have `Dual{<:AbstractFloat} <: AbstractFloat` so it loops endlessly.
I don't have a good solution for this at the moment. Maybe we need some dual array overloads too?

> Any concrete feedback on improving this @gdalle?

No, actually since `A \ b = pinv(A) * b` in most cases (is that true?), I'd say this is close to optimal.

> I'm planning to make heavy use of FillArrays.jl

This is a good idea.

adrhill:
Skipping `*` and `dot` in this PR, opened #133 to track.

Project.toml
```diff
@@ -25,6 +27,7 @@ SparseConnectivityTracerSpecialFunctionsExt = "SpecialFunctions"
 ADTypes = "1"
 Compat = "3,4"
 DocStringExtensions = "0.9"
+FillArrays = "1"
```
gdalle, line 30:
```suggestion
FillArrays = "1"
LinearAlgebra = "<0.0.1, 1"
```

test/runtests.jl
```diff
@@ -34,7 +34,10 @@ GROUP = get(ENV, "JULIA_SCT_TEST_GROUP", "Core")
                     Aqua.test_all(
                         SparseConnectivityTracer;
                         ambiguities=false,
-                        deps_compat=(ignore=[:Random, :SparseArrays], check_extras=false),
+                        deps_compat=(
+                            ignore=[:LinearAlgebra, :Random, :SparseArrays],
```
gdalle, line 38: I'd remove that

src/overloads/arrays.jl
```diff
@@ -0,0 +1,208 @@
...
+    primals = getproperty.(A, :primal)
+    tracers = getproperty.(A, :tracer)
+    return primals, tracers
+end
+function split_dual_array(A::SparseArrays.SparseMatrixCSC{D}) where {D<:Dual}
+    A = Matrix(A)
+    primals = getproperty.(A, :primal)
+    tracers = getproperty.(A, :tracer)
+    return sparse(primals), sparse(tracers)
+end
```
gdalle, line 66: why is this necessary?

src/overloads/arrays.jl
```diff
@@ -0,0 +1,208 @@
...
+#==================#
+
+# TODO: replace `second_order_or` by less conservative sparsity patterns when possible
+
+## Determinant
+LinearAlgebra.det(A::AbstractMatrix{T}) where {T<:AbstractTracer} = second_order_or(A)
+LinearAlgebra.logdet(A::AbstractMatrix{T}) where {T<:AbstractTracer} = second_order_or(A)
+function LinearAlgebra.logabsdet(A::AbstractMatrix{T}) where {T<:AbstractTracer}
+    t = second_order_or(A)
+    return (t, t)
```
gdalle, line 79: The second output of `logabsdet` is the sign of `det(A)`, so we can do better here: first and second derivatives are zero almost everywhere (only connectivity matters)

src/overloads/arrays.jl
```diff
@@ -0,0 +1,208 @@
...
+end
+
+#===========#
+# Utilities #
+#===========#
+
+function split_dual_array(A::AbstractArray{D}) where {D<:Dual}
+    primals = getproperty.(A, :primal)
+    tracers = getproperty.(A, :tracer)
+    return primals, tracers
```
gdalle, line 59: I prefer this variable naming to the `P, T` you use below (which sounds more like primal and tracer _types_)

src/overloads/arrays.jl
gdalle:
```suggestion
    return (t, sign(t))
```

src/overloads/arrays.jl
```diff
@@ -0,0 +1,208 @@
...
+## Norm
+function LinearAlgebra.norm(A::AbstractArray{T}, p::Real=2) where {T<:AbstractTracer}
+    return second_order_or(A)
+end
+function LinearAlgebra.opnorm(A::AbstractArray{T}, p::Real=2) where {T<:AbstractTracer}
+    return first_order_or(A)
+end
+function LinearAlgebra.opnorm(A::AbstractMatrix{T}, p::Real=2) where {T<:AbstractTracer}
+    return first_order_or(A)
+end
```
gdalle, line 111: this is wrong, in general the dependency is second-order (except for specific values of `p`)

src/overloads/arrays.jl
```diff
@@ -0,0 +1,208 @@
...
+
+## Determinant
+LinearAlgebra.det(A::AbstractMatrix{T}) where {T<:AbstractTracer} = second_order_or(A)
+LinearAlgebra.logdet(A::AbstractMatrix{T}) where {T<:AbstractTracer} = second_order_or(A)
+function LinearAlgebra.logabsdet(A::AbstractMatrix{T}) where {T<:AbstractTracer}
+    t = second_order_or(A)
+    return (t, t)
+end
+
+# Fix for issue #108
```
gdalle, line 82: I don't see why the stuff below is necessary?

src/overloads/arrays.jl
```diff
@@ -0,0 +1,208 @@
...
+## Exponent
+function LinearAlgebra.exp(A::AbstractMatrix{T}) where {T<:AbstractTracer}
+    LinearAlgebra.checksquare(A)
+    n = size(A, 1)
+    t = second_order_or(A)
+    return Fill(t, n, n)
+end
+
+## Matrix power
+function LinearAlgebra.:^(A::AbstractMatrix{T}, p::Integer) where {T<:AbstractTracer}
```
gdalle, line 170: here it might be worth specializing on the value of `p`, I assume powers `0` and `1` are common

src/overloads/arrays.jl
```diff
@@ -0,0 +1,208 @@
...
+#==============#
+
+# Conversion of matrices of tracers to SparseMatrixCSC has to be rewritten 
+# due to use of `count(_isnotzero, M)` in SparseArrays.jl
+#
+# Code modified from MIT licensed SparseArrays.jl source:
+# https://github.com/JuliaSparse/SparseArrays.jl/blob/45dfe459ede2fa1419e7068d4bda92d9d22bd44d/src/sparsematrix.jl#L901-L920
+# Copyright (c) 2009-2024: Jeff Bezanson, Stefan Karpinski, Viral B. Shah, and other contributors: https://github.com/JuliaLang/julia/contributors
+function SparseArrays.SparseMatrixCSC{Tv,Ti}(
+    M::StridedMatrix{Tv}
```
gdalle, line 188: why `Strided`?

src/overloads/arrays.jl
gdalle: what fails if you remove it?

test/test_arrays.jl
```diff
@@ -0,0 +1,122 @@
...
+using Test
+
+PATTERN_FUNCTIONS = (connectivity_pattern, jacobian_pattern, hessian_pattern)
+
+TEST_SQUARE_MATRICES = Dict(
+    "`Matrix` (3×3)" => rand(3, 3),
+    "`Symmetric` (3×3)" => Symmetric(rand(3, 3)),
+    "`Diagonal` (3×3)" => Diagonal(rand(3)),
+)
+TEST_MATRICES = merge(TEST_SQUARE_MATRICES, Dict("`Matrix` (3×4)" => rand(3, 4)))
```
gdalle, line 17: add sparse test matrices?

test/test_arrays.jl
```diff
@@ -0,0 +1,122 @@
...
+        @testset "`SparseMatrixCSC` (3×3)" begin
+            @test all(isone, connectivity_pattern(a -> opnorm(sparse(a), 1), rand(3, 3)))
+            @test all(isone, jacobian_pattern(a -> opnorm(sparse(a), 1), rand(3, 3)))
+            @test all(iszero, hessian_pattern(a -> opnorm(sparse(a), 1), rand(3, 3)))
+
+            @test all(isone, connectivity_pattern(a -> opnorm(spdiagm(a), 1), rand(3)))
+            @test all(isone, jacobian_pattern(a -> opnorm(spdiagm(a), 1), rand(3)))
+            @test all(iszero, hessian_pattern(a -> opnorm(spdiagm(a), 1), rand(3)))
+        end
+    end
```
gdalle, line 57: modify this because the detection is wrong

test/test_arrays.jl
```diff
@@ -0,0 +1,122 @@
...
+end
+
+@testset "Matrix-valued functions" begin
+    # Functions that only work on square matrices
+    @testset "$f" for f in (inv, exp, A -> A^3)
```
gdalle, line 62: add `^0` and `^1` with special behaviors

test/test_arrays.jl
```diff
@@ -0,0 +1,122 @@
...
+        sumf(x) = sum(f(x))
+        @testset "$name" for (name, A) in TEST_MATRICES
+            test_full_patterns(sumf, A)
+        end
+        @testset "`SparseMatrixCSC` (3×4)" begin
+            test_full_patterns(A -> sumf(sparse(A)), rand(3, 4))
+        end
+    end
+end
+
```
gdalle, line 85: what goes on below?

gdalle (changes requested):

gdalle:
I'd wait until we settle #135 before merging this one, especially due to the use of `myempty`

test/test_arrays.jl
adrhill: By overloading `literal_pow`?

src/overloads/arrays.jl
adrhill: To split a sparse matrix of duals into 
1. a sparse matrix of primals 
2. a sparse matrix of tracers

`split_dual_array` is called in `det`, `logdet` and `logabsdet`.

src/overloads/arrays.jl
adrhill: The comment refers to issue #108.

`det`, `logdet` and `logabsdet` fail on local sparsity detection without this:


```Julia
julia> using ADTypes, SparseConnectivityTracer

julia> using LinearAlgebra

julia> detector = TracerLocalSparsityDetector();

julia> g(x) = logdet(spdiagm(x))
g (generic function with 1 method)

julia> ADTypes.hessian_sparsity(g, randn(3), detector)
ERROR: StackOverflowError:
Stacktrace:
  ...
 [12] lu(A::SparseMatrixCSC{SparseConnectivityTracer.Dual{…}, Int64}; check::Bool) (repeats 18547 times)
    @ SparseArrays.UMFPACK ~/.julia/juliaup/julia-1.10.3+0.x64.linux.gnu/share/julia/stdlib/v1.10/SparseArrays/src/solvers/umfpack.jl:395
```

src/overloads/arrays.jl
adrhill: That's what LinearAlgebra.jl uses, so we need to subtype `StridedMatrix{Tv}`, `Tv<:AbstractTracer`.

test/test_arrays.jl
adrhill: Tests that are a bit more involved than testing for "full patterns".

test/test_arrays.jl
gdalle: Otherwise we can't test that the fixes you introduce actually solve #108

src/overloads/arrays.jl
gdalle: I guess I feel that this pattern of splitting the `Dual` into primal + tracer is very generic. We do it in our existing overloads for scalar-to-scalar operators, but maybe this means we should do the same for array-to-something operators?

test/test_arrays.jl
gdalle: no, just by checking the value of `p` in the source code that overloads `^` for tracer matrices. And by adding the functions `A -> A^0` and `A -> A^1` to the tests

test/test_arrays.jl
adrhill: No, that's what the closures in the tests are currently for.

src/overloads/arrays.jl
adrhill: Adding a comment on `lu`.

src/overloads/arrays.jl
```diff
@@ -0,0 +1,231 @@
...
+## Exponent
+function LinearAlgebra.exp(A::AbstractMatrix{T}) where {T<:AbstractTracer}
+    LinearAlgebra.checksquare(A)
+    n = size(A, 1)
+    t = second_order_or(A)
+    return Fill(t, n, n)
+end
+
+## Matrix power
+function LinearAlgebra.:^(A::AbstractMatrix{T}, p::Integer) where {T<:AbstractTracer}
```
gdalle, line 159: Is it a good idea to make this type-unstable? I don't know whether it matters, and I know I'm the one who asked for special cases

gdalle (changes requested):

src/overloads/arrays.jl
adrhill: We could remove the `isone` case.
I highly doubt it is relevant in the first place.

gdalle: I could see the case where people do something like `sum(A^k for k in 1:n)` but sure, let's remove it for now

adrhill: Well, in that very specific case, were also more efficient now, unless `n=1`. ;)

PR #132, opened 2024-06-19
Remove overloads on `similar`

### LINKED ISSUE #123: Remove overloads on `similar`
These are currently only required for NNlib compatibility with convolutional layers, as the following error is otherwise thrown:

```julia
 TaskFailedException
  
      nested task error: UndefRefError: access to undefined reference
      Stacktrace:
       [1] getindex
         @ ./essentials.jl:14 [inlined]
       [2] getindex
         @ ./subarray.jl:290 [inlined]
       [3] conv_direct!(y::SubArray{Dual{Float64, GradientTracer{DuplicateVector{Int64}}}, 5, Array{Dual{Float64, GradientTracer{DuplicateVector{Int64}}}, 5}, Tuple{Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, UnitRange{Int64}, Base.Slice{Base.OneTo{Int64}}}, false}, x::SubArray{Dual{Float64, GradientTracer{DuplicateVector{Int64}}}, 5, Array{Dual{Float64, GradientTracer{DuplicateVector{Int64}}}, 5}, Tuple{Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, UnitRange{Int64}, Base.Slice{Base.OneTo{Int64}}}, false}, w::SubArray{Float64, 5, Array{Float64, 5}, Tuple{Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, UnitRange{Int64}}, true}, cdims::NNlib.DenseConvDims{3, 3, 3, 6, 3}, ::Val{(2, 2, 1)}, ::Val{1}, ::Val{(0, 0, 0, 0, 0, 0)}, ::Val{(1, 1, 1)}, ::Val{(1, 1, 1)}, fk::Val{false}; alpha::Dual{Float64, GradientTracer{DuplicateVector{Int64}}}, beta::Bool)
         @ NNlib ~/.julia/packages/NNlib/g1Wwt/src/impl/conv_direct.jl:111
       [4] conv_direct!(y::SubArray{Dual{Float64, GradientTracer{DuplicateVector{Int64}}}, 5, Array{Dual{Float64, GradientTracer{DuplicateVector{Int64}}}, 5}, Tuple{Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, UnitRange{Int64}, Base.Slice{Base.OneTo{Int64}}}, false}, x::SubArray{Dual{Float64, GradientTracer{DuplicateVector{Int64}}}, 5, Array{Dual{Float64, GradientTracer{DuplicateVector{Int64}}}, 5}, Tuple{Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, UnitRange{Int64}, Base.Slice{Base.OneTo{Int64}}}, false}, w::SubArray{Float64, 5, Array{Float64, 5}, Tuple{Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, UnitRange{Int64}}, true}, cdims::NNlib.DenseConvDims{3, 3, 3, 6, 3}; alpha::Dual{Float64, GradientTracer{DuplicateVector{Int64}}}, beta::Bool)
         @ NNlib ~/.julia/packages/NNlib/g1Wwt/src/impl/conv_direct.jl:50
       [5] conv_direct!
         @ ~/.julia/packages/NNlib/g1Wwt/src/impl/conv_direct.jl:47 [inlined]
       [6] (::NNlib.var"#307#311"{@Kwargs{}, NNlib.DenseConvDims{3, 3, 3, 6, 3}, SubArray{Dual{Float64, GradientTracer{DuplicateVector{Int64}}}, 5, Array{Dual{Float64, GradientTracer{DuplicateVector{Int64}}}, 5}, Tuple{Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, UnitRange{Int64}, Base.Slice{Base.OneTo{Int64}}}, false}, SubArray{Float64, 5, Array{Float64, 5}, Tuple{Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, UnitRange{Int64}}, true}, SubArray{Dual{Float64, GradientTracer{DuplicateVector{Int64}}}, 5, Array{Dual{Float64, GradientTracer{DuplicateVector{Int64}}}, 5}, Tuple{Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, Base.Slice{Base.OneTo{Int64}}, UnitRange{Int64}, Base.Slice{Base.OneTo{Int64}}}, false}})()
         @ NNlib ~/.julia/packages/NNlib/g1Wwt/src/conv.jl:209
```

However, this does look like a workaround for unusual behavior of NNlib, since the following line most likely shouldn't try to access `undef` variables:
https://github.com/FluxML/NNlib.jl/blob/381a41f2b912eb924f1a6c256a9bf49a593e4b67/src/impl/conv_direct.jl#L111

adrhill:
Closes #123.

### DIFF
--- a/src/overloads/conversion.jl
+++ b/src/overloads/conversion.jl
@@ -24,20 +24,6 @@ Base.floatmin(::Type{T})    where {T<:AbstractTracer} = myempty(T)
 Base.floatmax(::Type{T})    where {T<:AbstractTracer} = myempty(T)
 Base.maxintfloat(::Type{T}) where {T<:AbstractTracer} = myempty(T)
 
-## Array constructors
-# TODO: remove (#123)
-Base.similar(a::Array{T,1})                     where {T<:AbstractTracer}   = zeros(T, size(a, 1))
-Base.similar(a::Array{T,2})                     where {T<:AbstractTracer}   = zeros(T, size(a, 1), size(a, 2))
-Base.similar(a::Array{A,1}, ::Type{T})          where {T<:AbstractTracer,A} = zeros(T, size(a, 1))
-Base.similar(a::Array{A,2}, ::Type{T})          where {T<:AbstractTracer,A} = zeros(T, size(a, 1), size(a, 2))
-Base.similar(::Array{T}, m::Int)                where {T<:AbstractTracer}   = zeros(T, m)
-
-# These have to be defined on specific types to exclude `Union{}` triggered by Documenter.jl
-# TODO: remove (#123)
-Base.similar(::Array, ::Type{ConnectivityTracer{I}}, dims::Dims{N}) where {I,N}   = zeros(ConnectivityTracer{I}, dims)
-Base.similar(::Array, ::Type{GradientTracer{G}},     dims::Dims{N}) where {G,N}   = zeros(GradientTracer{G}, dims)
-Base.similar(::Array, ::Type{HessianTracer{G,H}},    dims::Dims{N}) where {G,H,N} = zeros(HessianTracer{G,H}, dims)
-
 ## Duals
 function Base.promote_rule(::Type{Dual{P1, T}}, ::Type{Dual{P2, T}}) where {P1,P2,T}
     PP = Base.promote_type(P1, P2) # TODO: possible method call error?
@@ -75,40 +61,5 @@ Base.eps(::Type{D})         where {P,T,D<:Dual{P,T}} = D(eps(P),         myempty
 Base.floatmin(::Type{D})    where {P,T,D<:Dual{P,T}} = D(floatmin(P),    myempty(T))
 Base.floatmax(::Type{D})    where {P,T,D<:Dual{P,T}} = D(floatmax(P),    myempty(T))
 Base.maxintfloat(::Type{D}) where {P,T,D<:Dual{P,T}} = D(maxintfloat(P), myempty(T))
-
-## Array constructors
-# TODO: remove (#123)
-function Base.similar(a::Array{D,1}) where {P,T,D<:Dual{P,T}}
-    p_out = similar(primal.(a))
-    return Dual.(p_out, myempty(T))
-end
-function Base.similar(a::Array{D,2}) where {P,T,D<:Dual{P,T}}
-    p_out = similar(primal.(a))
-    return Dual.(p_out, myempty(T))
-end
-function Base.similar(a::Array{A,1}, ::Type{D}) where {P,T,D<:Dual{P,T},A}
-    p_out = similar(a, P)
-    return Dual.(p_out, myempty(T))
-end
-function Base.similar(a::Array{A,2}, ::Type{D}) where {P,T,D<:Dual{P,T},A}
-    p_out = similar(a, P)
-    return Dual.(p_out, myempty(T))
-end
-function Base.similar(a::Array{D}, m::Int) where {P,T,D<:Dual{P,T}}
-    p_out = similar(primal.(a), m)
-    return Dual.(p_out, myempty(T))
-end
-function Base.similar(a::Array{D}, dims::Dims{N}) where {P,T,D<:Dual{P,T}, N}
-    p_out = similar(primal.(a), dims)
-    return Dual.(p_out, myempty(T))
-end
-function Base.similar(a::Array{D2}, ::Type{Dual{P,T}}, dims::Dims{N}) where {P,T,N,D2<:Dual}
-    p_out = similar(primal.(a), P, dims)
-    return Dual.(p_out, myempty(T))
-end
-function Base.similar(a::Array, ::Type{Dual{P,T}}, dims::Dims{N}) where {P,T,N}
-    p_out = similar(a, P, dims)
-    return Dual.(p_out, myempty(T))
-end
 
 #! format: on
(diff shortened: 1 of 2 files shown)


adrhill:
Looks like NNlib `v0.9.X` only supports Julia 1.9 and up.

adrhill:
The way I see it, we have two options for Julia 1.6 support:
* we overload `similar` in SCT on Julia <1.9
* we overload `similar` in the NNlib extension on NNlib <0.9.18
* we use the upcoming SCT `v0.6` release to wipe our slate clean and break compatibility of Julia <1.9 with `conv_direct!`

I'm in favor of the last option. How about you @gdalle?

gdalle:
I favor the first option, because the whole NLP ecosystem wants to support Julia 1.6, and now ADNLPModels relies on SCT.

adrhill:
The overloads on `similar` are only required for `conv_direct!` in NNlib.
The NLP ecosystem is unrelated, as you can see by the tests passing.

gdalle:
Oh right, sorry. Then yeah third option

ErikQQY:
Hi, I recently encountered a scenario where I need to do `similar` on `SCT.Dual`, I may not be very familiar with the context here, but why should we remove the overloads on similar? Without the overloads, `similar(Vector{SCT.Dual})` would directly throw `#undef` which is pretty annoying.🤔

adrhill:
> Hi, I recently encountered a scenario where I need to do `similar` on `SCT.Dual`

Could you give me more information on this? Is the code you want to trace calling `similar` or do you want to allocate memory? 

For the latter, we've recently (#232) added a new stable API: https://adrianhill.de/SparseConnectivityTracer.jl/stable/user/api/#Memory-allocation

> why should we remove the overloads on similar? Without the overloads, `similar(Vector{SCT.Dual})` would directly throw `#undef`

I might be wrong here, but isn't this the expected behavior of `similar`?

> ```
> help?> similar
> search: similar
>
>   similar(array, [element_type=eltype(array)], [dims=size(array)])
>
>   Create an uninitialized mutable array with the given element type and size, based upon the given source array. [...]
> ```

gdalle:
The general idea is that if you access an array element initialized by similar before having written anything to it, that's a problem with your code

adrhill:
`similar` is expected to return unitialized arrays and commonly does:

```Julia-repl
julia> sets = [Set(i) for i in 1:5]
5-element Vector{Set{Int64}}:
 Set([1])
 Set([2])
 Set([3])
 Set([4])
 Set([5])

julia> similar(sets)
5-element Vector{Set{Int64}}:
 #undef
 #undef
 #undef
 #undef
 #undef
```

Using the outputs of `similar` in a subsequent computation sounds like a bug to me. Maybe your code should call `zeros` instead?

> curious why the similar `overloads` are removed.

Our `similar` used to call `zeros` for compatibility with NNlib, which used to have a similar bug where undefined values were accessed. This bug has since been fixed in NNlib, hence #123.

There is a difference in performance between returning `zeros` or an uninitialized array:

```Julia-repl
julia> T
SparseConnectivityTracer.GradientTracer{SparseConnectivityTracer.IndexSetGradientPattern{Int64, BitSet}}

julia> A = rand(1000, 1000);

julia> @benchmark similar($A, $T)
BenchmarkTools.Trial: 9795 samples with 1 evaluation per sample.
 Range (min … max):  388.750 μs …   1.780 ms  ┊ GC (min … max):  0.00% … 13.75%
 Time  (median):     419.250 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   509.660 μs ± 143.131 μs  ┊ GC (mean ± σ):  16.09% ± 17.15%

  █▂▃▅▂▁▁ ▁   ▂▁           ▂   ▃▆▃▃▄▁▁      ▁                   ▁
  ██████████████▇▇▇▆▇▆█▇██▇███████████████▇▇██▆▆▆▆▆▆▇▇▆▇▆▇▆▇▆▆▆ █
  389 μs        Histogram: log(frequency) by time        887 μs <

 Memory estimate: 15.26 MiB, allocs estimate: 3.

julia> @benchmark zeros($T, $1000, $1000)
BenchmarkTools.Trial: 5538 samples with 1 evaluation per sample.
 Range (min … max):  746.250 μs …   2.284 ms  ┊ GC (min … max):  0.00% … 60.50%
 Time  (median):     964.770 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   901.104 μs ± 148.584 μs  ┊ GC (mean ± σ):  11.88% ± 10.85%

  █▄▄▅▂▁     ▁           ▂▇▅▅▅▃▂▁▁▁  ▁▁                         ▁
  █████████████▇▇▆▆▅▅▆▆▇▇███████████████▆█▇▅▇▆▆▇▇▇▇▇▆▇▆▇▆▆▆▃▄▂▅ █
  746 μs        Histogram: log(frequency) by time        1.3 ms <

 Memory estimate: 15.26 MiB, allocs estimate: 8.
```

ErikQQY:
Oh I see, it's indeed a problem with our code then

PR #135, opened 2024-06-24
Add shared Hessian tracer à la Walther

gdalle:
**Source**

- [x] Add the `shared` type parameter to `IndexSetHessianPattern`, signifying that the `hessian` field is shared between all scalar quantities
- [x] Modify `1-to-1` and `2-to-1` Hessian overloads to leverage shared `hessian` fields in `IndexSetHessianPattern` and avoid creating new objects
- [x] Add a `create_pattern` routine which is needed for the creation of shared patterns
- [x] Handle special case overloads (`ifelse` and the like)

**Tests**

- [x] Add one shared `IndexSetHessianPattern` to the test loops
- [x] Adjust tests to take dead ends into account when the `IndexSetHessianPattern` is shared

**Questions**

Should `myempty` be defined for `IndexSetHessianPattern{I,SG,SH,true}`?

- NO: The only meaningful Hessian tracer will be constructed at the start by `create_tracers`, and if we make another it means we screwed up somewhere and it may result in incorrect behavior / bad performance.
- YES: I _think_ it's correct (but still suboptimal performance-wise) to have empty tracers appear along the function, because the overloads take care to ensure consistency between tracers through this line
https://github.com/adrhill/SparseConnectivityTracer.jl/blob/3a1df201c6864d50b73359473db058df9b8271d2/src/overloads/hessian_tracer.jl#L116


adrhill:
Awesome! Did you already compare the performance to existing set types?

gdalle:
> Awesome! Did you already compare the performance to existing set types?

No, nor did I include it in the benchmarks.

Depending on what we decide for `myempty`, the performance might be rather different.

src/overloads/hessian_tracer.jl
```diff
@@ -7,24 +7,32 @@
...
     end
 end
 
 function hessian_tracer_1_to_1_inner(
-    sg::G, sh::H, is_der1_zero::Bool, is_secondder_zero::Bool
+    sg::G, sh::H, is_der1_zero::Bool, is_secondder_zero::Bool; shared::Bool
 ) where {I<:Integer,G<:AbstractSet{I},H<:AbstractSet{Tuple{I,I}}}
     sg_out = gradient_tracer_1_to_1_inner(sg, is_der1_zero)
-    sh_out = if is_der1_zero && is_secondder_zero
-        myempty(H)
```
adrhill, line 21: Why did you remove this case?

test/test_hessian.jl
```diff
@@ -99,20 +101,38 @@ HESSIAN_TRACERS = (
...
+                0 0 1 0
+            ]
+        else
+            @test h == [
+                0 0 0 0
+                0 0 0 0
+                0 0 0 0
+                0 0 0 0
+            ]
+        end
```
adrhill, line 135: I think we could return a fully sparse pattern here, even with in-place mutation.

adrhill: See here: https://github.com/adrhill/SparseConnectivityTracer.jl/pull/135/files/3a1df201c6864d50b73359473db058df9b8271d2#diff-438bb091bbff5d5dec5b45b3c61577893cbbd4438cc8fd853135986a34bd8443

adrhill:
> Should `myempty` be defined for `HessianTracer{G,H,true}`?

For mutable sets absolutely. 

> NO: The only meaningful Hessian tracer will be constructed at the start by `create_tracers`, and if we make another it means we screwed up somewhere and it may result in incorrect behavior / bad performance.

I think carrying around smaller sets instead of needlessly large ones could actually increase the performance of the tracing. Needless to say, the sparse AD pass will also be much faster due to increased sparsity.

src/tracers.jl
```diff
@@ -200,13 +206,16 @@ end
 #===========#
 
 myempty(::Type{ConnectivityTracer{I}}) where {I} = ConnectivityTracer{I}(myempty(I), true)
-myempty(::Type{GradientTracer{G}}) where {G}     = GradientTracer{G}(myempty(G), true)
-myempty(::Type{HessianTracer{G,H}}) where {G,H}  = HessianTracer{G,H}(myempty(G), myempty(H), true)
+myempty(::Type{GradientTracer{G}}) where {G} = GradientTracer{G}(myempty(G), true)
+
+function myempty(::Type{HessianTracer{G,H,shared}}) where {G,H,shared}
+    return HessianTracer{G,H,shared}(myempty(G), myempty(H), true)
+end
```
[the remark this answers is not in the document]
adrhill: Ok, I see how that's equivalent if you never create new sets.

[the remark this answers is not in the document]
gdalle: Yes it is globally shared, because of the way I modified tracer creation

src/tracers.jl
```diff
@@ -218,8 +227,45 @@ end
...
+
+function create_tracers(
+    ::Type{Dual{P,HessianTracer{G,H,true}}},
+    xs::AbstractArray{<:Real,N},
+    indices::AbstractArray{<:Integer,N},
+) where {P<:Real,G,H,N}
+    sh = myempty(H)  # shared
+    return map(xs, indices) do x, index
+        Dual(x, HessianTracer{G,H,true}(seed(G, index), sh))
+    end
```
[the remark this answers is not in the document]
adrhill: Thanks, got it now! :)

src/tracers.jl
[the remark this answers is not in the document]
gdalle: It is also equivalent if all you create are new empty sets with no relation to the initial indices

src/tracers.jl
```diff
@@ -117,30 +117,36 @@ $(TYPEDEF)
 
 For a higher-level interface, refer to [`hessian_pattern`](@ref).
 
+The last type parameter `shared` is a `Bool` indicating whether the Hessian should be shared among all intermediate scalar quantities.
+
 ## Fields
 $(TYPEDFIELDS)
 """
-struct HessianTracer{G,H} <: AbstractTracer
+struct HessianTracer{G,H,shared} <: AbstractTracer
```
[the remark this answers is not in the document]
gdalle: The way the code is currently written also assumes mutability of `G` and `H`. And it assumes they are compatible with `union_product!(::H, ::G, ::G)`. There are plenty of things we cannot enforce statically.

I don't think patterns should be revisited, because they add an extra layer of complexity. But if you do want to give them another go, please wait until I have opened the "array storage" PR (as opposed to "set storage") in order to have concrete examples.

gdalle:
> For mutable sets absolutely.

We don't have a static handle on which set types are mutable and which are not.

> I think carrying around smaller sets instead of needlessly large ones could actually increase the performance of the tracing. 

You might be misunderstanding my question. I'm not saying we should get rid of the "non-Walther" mode. I'm saying that in the "Walther" (shared) mode, we still allow the creation of empty Hessians, and I'm not sure it's necessary or even a good idea.

test/test_hessian.jl
gdalle: No we cannot. In shared mode, the resulting pattern is the union of all patterns of intermediate quantities. `x[1] * x[2]` and `x[3] * x[4]` are intermediate quantities

src/tracers.jl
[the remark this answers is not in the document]
adrhill: Yes, don't get me wrong, we'll merge this without any larger detours.

> please wait until I have opened the "array storage" PR

Sure!

adrhill:
Sorry, my review and my comment on `myempty` were written before understanding the role `create_tracers` plays (https://github.com/adrhill/SparseConnectivityTracer.jl/pull/135#discussion_r1650754725). The crux is that only one tracer is ever created, indeed making them analogous to Walther's approach.

This does make the question on empty tracer construction interesting. Maybe the solution is to add methods

```Julia
myempty(h::HessianTracer{G,H,true}) where {G, H} = t
myempty(::T) where {T<:AbstractTracer} = myempty(T)
```

if we truly just want to have a single tracer instance.

src/tracers.jl
```diff
@@ -200,13 +206,16 @@ end
 #===========#
 
 myempty(::Type{ConnectivityTracer{I}}) where {I} = ConnectivityTracer{I}(myempty(I), true)
-myempty(::Type{GradientTracer{G}}) where {G}     = GradientTracer{G}(myempty(G), true)
-myempty(::Type{HessianTracer{G,H}}) where {G,H}  = HessianTracer{G,H}(myempty(G), myempty(H), true)
+myempty(::Type{GradientTracer{G}}) where {G} = GradientTracer{G}(myempty(G), true)
```
adrhill, line 209: Why not support shared `Gradient`- and `ConnectivityTracer`s?

test/test_hessian.jl
```diff
@@ -1,15 +1,17 @@
 using SparseConnectivityTracer
-using SparseConnectivityTracer: Dual, HessianTracer, MissingPrimalError, trace_input, empty
+using SparseConnectivityTracer:
+    Dual, HessianTracer, MissingPrimalError, trace_input, empty, isshared
```
adrhill, line 3: Can we add a test where we manually create a shared tracer via `create_tracers`, run it through a function with multiple outputs and check for egality of the outputs with `===`? This could ensure we don't accidentally create new tracers somewhere.

test/test_hessian.jl
```diff
@@ -1,15 +1,17 @@
...
 HESSIAN_TRACERS = (
-    HessianTracer{BitSet,Set{Tuple{Int,Int}}},
-    HessianTracer{Set{Int},Set{Tuple{Int,Int}}},
-    HessianTracer{DuplicateVector{Int},DuplicateVector{Tuple{Int,Int}}},
-    HessianTracer{SortedVector{Int},SortedVector{Tuple{Int,Int}}},
+    HessianTracer{BitSet,Set{Tuple{Int,Int}},false},
+    HessianTracer{BitSet,Set{Tuple{Int,Int}},true},
+    HessianTracer{Set{Int},Set{Tuple{Int,Int}},false},
+    HessianTracer{DuplicateVector{Int},DuplicateVector{Tuple{Int,Int}},false},
+    HessianTracer{SortedVector{Int},SortedVector{Tuple{Int,Int}},false},
```
adrhill, line 14: A similar change has to be made to our benchmark suite.

src/tracers.jl
gdalle: What's interesting about the shared Hessian is that the gradient part of each `HessianTracer` can still vary

test/test_hessian.jl
gdalle: I don't see how that would safeguard against new tracer creation. And such a test is likely subsumed by the ones I added in the Hessian pattern section

gdalle: I added tests with the function below, but here are a few caveats.

```julia
function multi_output_for_shared_test(x::AbstractArray)
    z = ones(eltype(x), size(x))
    y1 = x[1]^2 * z[1]
    y2 = z[2] * x[2]^2
    y3 = x[1] * x[2]
    y4 = z[1] * z[2]
    return [y1, y2, y3, y4]
end
```

- When two `HessianTracer`s are merged in "shared" mode, it's arbitrary which one gets picked to be mutated because they're _supposed_ to be the same object. So in theory, `y1` and `y2` could end up being different objects.
  - It doesn't happen here because of the emptiness filter, which is applied before the `_inner` routine: `z[1]` and `z[2]` are empty so they're not even considered.
  - It could happen theoretically if a non-empty tracer were created with no relation whatsoever to the initial array. However I'm not sure that's a state our code could ever be in.
- Since the contents of `z` are new tracer objects, the last coordinate `y4` contains a different Hessian object than the rest. I'm not sure that can be solved, except by disallowing `myempty` in "shared" mode as I suggested. Hence the following tests:

```julia
yt = # output of the function on a shared HessianTracer array
@test yt[1].hessian == yt[2].hessian
@test yt[1].hessian == yt[3].hessian
@test_broken yt[1].hessian == yt[4].hessian
```

test/test_hessian.jl
gdalle: I took a middle ground and added a global benchmark for shared mode

src/tracers.jl
adrhill: So in #119 lingo, you have a `SetVectorPattern` and a `SharedMatrixPattern`. 😉

I hope you see the point of the added pattern abstraction: it's a nuisance to make these changes on the three tracer types in almost every PR.

gdalle: Why would we ever need a shared anything for `ConnectivityTracer` or `GradientTracer`?
If you share the underlying storage, every output variable has the same gradient pattern, so your Jacobian has constant rows (not columns, sorry).

gdalle:
Don't merge, the benchmarks show an epic slowdown in the shared version compared to the individual one. I'll investigate

src/tracers.jl
```diff
@@ -117,30 +117,36 @@ $(TYPEDEF)
 
 For a higher-level interface, refer to [`hessian_pattern`](@ref).
 
+The last type parameter `shared` is a `Bool` indicating whether the Hessian should be shared among all intermediate scalar quantities.
```
adrhill, line 120: Could you emphasize that only the `hessian` field in `HessianTracer` is shared? 
The word "Hessian" is a bit ambiguous.

adrhill: I think that should be enough documentation until we decide to export shared pattern functionality.

It's going to be non-trivial to explain to users why they can obtain four different types of sparsity patterns.

adrhill:
> Don't merge, the benchmarks show an epic slowdown in the shared version compared to the individual one.

It also looks like there is a minor regression on non-shared Hessians. It could just be noise, but that looks like a pattern: all of them appear on `Conv`.

I don't consider this blocking, but it's worth investigating.

src/tracers.jl
adrhill: The same detail should be explained in the `create_tracers` docstring as well.

gdalle: I added a warning saying that `shared` is not part of the public API and should always be `false`. I'm not entirely happy with that because it means that users _do_ have to set it. But I think this is better than needing overloads for a partially-defined type `HessianTracer{G,H}`

adrhill: Yeah, I prefer enforcing full type definitions as well.

gdalle:
I took a look at the flame graph for one of the instances where runtime and memory consumption exploded in shared mode. There are no unexpected copies, in fact the total number of allocations has gone down as expected. However, we spend the majority of the time in `push!` and `sizehint!` for the Hessian tracer: those operations are more expensive because each 2d set involved is much bigger.

I suggest we keep this PR on the backburner for now, I'll open another one for a sparse matrix representation and we'll see whether that improves things. I'll also try to think about clever representations for a 2d set that only grows and never shrinks.

adrhill:
> I'll also try to think about clever representations for a 2d set that only grows and never shrinks.

Depending on the problem, it might actually be reasonable to use "dense" BitVectors and BitMatrices.

adrhill:
I'd like to see this merged soon. 
It's a correct implementation of an undocumented feature and adds useful tests. The performance can be improved in a follow-up.

src/overloads/hessian_tracer.jl
```diff
@@ -13,22 +13,36 @@ end
...
+) where {I,SG,SH,P<:IndexSetHessianPattern{I,SG,SH,false}}
     sg = gradient(p)
     sh = hessian(p)
     sg_out = gradient_tracer_1_to_1_inner(sg, is_der1_zero)
     sh_out = if is_der1_zero && is_der2_zero
         myempty(SH)
     elseif !is_der1_zero && is_der2_zero
         sh
-    elseif is_der1_zero && !is_der2_zero
+    elseif is_der1_zero && is_der2_zero
```
gdalle, line 24: missing a `!` here and it wasn't caught by the tests

src/patterns.jl
```diff
@@ -14,6 +14,14 @@ AbstractPattern
 """
 abstract type AbstractPattern end
 
+"""
+    isshared(pattern)
+
+Indicates whether patterns share memory (mutate).
```
gdalle, line 20: Not the same concept. Our non-shared patterns are still mutating when we perform operations like `union!`

src/patterns.jl
```diff
@@ -118,29 +124,45 @@ For use with [`HessianTracer`](@ref).
...
+function myempty(::Type{P}) where {I,SG,SH,M,P<:IndexSetHessianPattern{I,SG,SH,M}}
+    return P(myempty(SG), myempty(SH))
 end
-function seed(::Type{IndexSetHessianPattern{I,SG,SH}}, index) where {I,SG,SH}
-    return IndexSetHessianPattern{I,SG,SH}(seed(SG, index), myempty(SH))
+function create_patterns(
+    ::Type{P}, xs, is
+) where {I,SG,SH,M,P<:IndexSetHessianPattern{I,SG,SH,M}}
+    gradients = seed.(SG, is)
+    hessian = myempty(SH)
```
gdalle, line 164: If `M == false` we still want one hessian per input!

gdalle: Let's activate benchmarks on this PR, in your haste this would have uniformly tanked performance beyond belief

src/patterns.jl
```diff
@@ -118,29 +124,45 @@ For use with [`HessianTracer`](@ref).
...
 """
-struct IndexSetHessianPattern{I<:Integer,SG<:AbstractSet{I},SH<:AbstractSet{Tuple{I,I}}} <:
-       AbstractHessianPattern
+struct IndexSetHessianPattern{
+    I<:Integer,SG<:AbstractSet{I},SH<:AbstractSet{Tuple{I,I}},mutating
+} <: AbstractHessianPattern
     gradient::SG
     hessian::SH
 end
+isshared(::Type{IndexSetHessianPattern{I,SG,SH,true}}) where {I,SG,SH} = true
```
gdalle, line 155:
```suggestion
isshared(::Type{IndexSetHessianPattern{I,SG,SH,M}}) where {I,SG,SH,M} = M
```

src/patterns.jl
```diff
@@ -118,29 +124,45 @@ For use with [`HessianTracer`](@ref).
...
+
+## Internals
 
-Gradient and Hessian sparsity patterns constructed by combining two AbstractSets.
+The last type parameter `shared` is a `Bool` indicating whether the `hessian` field of this object should be shared among all intermediate scalar quantities involved in a function.
 """
-struct IndexSetHessianPattern{I<:Integer,SG<:AbstractSet{I},SH<:AbstractSet{Tuple{I,I}}} <:
-       AbstractHessianPattern
+struct IndexSetHessianPattern{
+    I<:Integer,SG<:AbstractSet{I},SH<:AbstractSet{Tuple{I,I}},mutating
```
gdalle, line 150: I prefer `shared` to `mutating`, see above

gdalle: Its defining property is that it is shared between indices

gdalle: Theoretically there could be a non mutating version which just copies it, if copies were free

src/patterns.jl
adrhill: Is the following more clear?

> Indicates whether patterns share memory, allowing operators to mutate their arguments.

> Our non-shared patterns are still mutating when we perform operations like union!

Internal allocations and mutations of said new allocations aren't usually what Julia developers refer to as `mutating functions`.

src/patterns.jl
adrhill: > If M == false we still want one hessian per input!

> this would have uniformly tanked performance beyond belief

Why would we want that? I think it increases performance.

src/patterns.jl
adrhill: I feel like the status quo might make it more obvious to the Julia compiler that `isshared` always returns a Boolean value (?)

src/patterns.jl
adrhill: I chose this because I feel like mutation is a much more common concept for Julia devs.
Walther's approach mutates values stored at a shared pointer.

And technically speaking, in this PR, pointers aren't *always* shared, see the `myempty` discussion above and https://github.com/adrhill/SparseConnectivityTracer.jl/pull/135#discussion_r1650967036. What this PR does is **allow** operators to mutate their tracer arguments.

src/patterns.jl
adrhill: Especially without mutation, we want to avoid allocations wherever possible.

src/patterns.jl
adrhill: I renamed it, but I hope you see that mutating operators are the defining concept of this PR.

src/patterns.jl
adrhill: See https://github.com/adrhill/SparseConnectivityTracer.jl/commit/c5e4f51af8c710804cfc63d36fbdcb176772d4ae for clarification.

src/patterns.jl
adrhill: See https://github.com/adrhill/SparseConnectivityTracer.jl/commit/c5e4f51af8c710804cfc63d36fbdcb176772d4ae

adrhill:
The `hessian_tracer_1_to_1_inner` call still needs meaningful tests on `is_der1_zero && !is_der2_zero` and `is_der1_zero && is_der2_zero` (https://github.com/adrhill/SparseConnectivityTracer.jl/pull/135#discussion_r1658223111).

src/overloads/hessian_tracer.jl
```diff
@@ -110,6 +126,29 @@ function hessian_tracer_2_to_1_inner(
...
+    is_der1_arg2_zero::Bool,
+    is_der2_arg2_zero::Bool,
+    is_der_cross_zero::Bool,
+) where {I,SG,SH,P<:IndexSetHessianPattern{I,SG,SH,true}}
+    sgx, shx = gradient(px), hessian(px)
+    sgy, shy = gradient(py), hessian(py)
+    shx !== shy && error("Expected shared Hessians, got $shx, $shy.")
+
+    sg_out = gradient_tracer_2_to_1_inner(sgx, sgy, is_der1_arg1_zero, is_der1_arg2_zero)
+    sh_out = union!(shx, shy)
```
gdalle, line 144: This is actually unnecessary since they are the same object

src/overloads/hessian_tracer.jl
```diff
@@ -110,6 +126,29 @@ function hessian_tracer_2_to_1_inner(
...
+    py::P,
+    is_der1_arg1_zero::Bool,
+    is_der2_arg1_zero::Bool,
+    is_der1_arg2_zero::Bool,
+    is_der2_arg2_zero::Bool,
+    is_der_cross_zero::Bool,
+) where {I,SG,SH,P<:IndexSetHessianPattern{I,SG,SH,true}}
+    sgx, shx = gradient(px), hessian(px)
+    sgy, shy = gradient(py), hessian(py)
+    shx !== shy && error("Expected shared Hessians, got $shx, $shy.")
```
gdalle, line 141: We're going the strict way, I like it

src/patterns.jl
```diff
@@ -118,29 +129,47 @@ For use with [`HessianTracer`](@ref).
...
-       AbstractHessianPattern
-    gradient::SG
-    hessian::SH
+struct IndexSetHessianPattern{
+    I<:Integer,G<:AbstractSet{I},H<:AbstractSet{Tuple{I,I}},shared
+} <: AbstractHessianPattern
+    gradient::G
+    hessian::H
 end
+isshared(::Type{IndexSetHessianPattern{I,G,H,true}}) where {I,G,H} = true
```
gdalle, line 160:
```suggestion
isshared(::Type{IndexSetHessianPattern{I,G,H,shared}}) where {I,G,H,shared} = shared
```

src/overloads/hessian_tracer.jl
adrhill: Does this case even exist?

gdalle: yes locally

adrhill:
> The `hessian_tracer_1_to_1_inner` call still needs meaningful tests on `is_der1_zero && !is_der2_zero` and `is_der1_zero && is_der2_zero` (https://github.com/adrhill/SparseConnectivityTracer.jl/pull/135#discussion_r1658223111).

I've added the missing test for `is_der1_zero && is_der2_zero`.
The required local overloads on to hit the `is_der1_zero && !is_der2_zero` case on 1-to-1 functions don't exist yet. I've added a comment documenting this.

PR #136, opened 2024-06-24
Put benchmark boilerplate into subpackage

gdalle:
Some code defining problem instances is used by tests and benchmarks, as well as external [notebooks](https://github.com/gdalle/SparsityDetectionComparison). Putting it in a package dispenses us from using `include` every time, and brings cleaner dependency management, at the cost of a `Pkg.develop`.

1. Create a subpackage in `benchmarks/SparseConnectivityTracerBenchmarks` to put:
  - the Brusselator code
  - the NLPModels code
  - possibly the optimal power flow code or other stuff in the future
2. Use [PkgJogger.jl](https://github.com/awadell1/PkgJogger.jl) to define the benchmark and run each `@benchmarkable` just once in the test suite


adrhill:
Isn't that what the `benchmark` folder is for? Why don't we move everything in there and add a `Project.toml`?

gdalle:
No, the benchmark folder is for benchmarking only, and has deps like BenchmarkTools. This code is used in tests, benchmarks and other places too. And it needs to be in a package to be importable in e.g. a Pluto notebook (to which I can give the exact GitHub URL of `libs/SparseConnectivityTracerBenchmarks`)

gdalle:
The name may be ill-chosen, perhaps `SparseConnectivityTracerProblems` would be better?

adrhill:
> No, the benchmark folder is for benchmarking only

Not trying to be snarky here, but isn't `libs/SparseConnectivityTracerBenchmarks` also about benchmarking?

> and has deps like BenchmarkTools

Is that a big issue? I think removing some complexity at the cost of one dependency is acceptable.

> to which I can give the exact GitHub URL of `libs/SparseConnectivityTracerBenchmarks`)

Wouldn't `benchmark` be a URL that's just as viable?

adrhill:
At the very least, we could move this into `benchmarks/libs/SparseConnectivityTracerBenchmarks` if this somehow is necessary for Pluto.

gdalle:
> Not trying to be snarky here, but isn't libs/SparseConnectivityTracerBenchmarks also about benchmarking?

See my latest comment, here I meant "benchmarks" in the sense of "benchmark instance definitions", not "full specification of a benchmark suite with sizes and evals and so on". Hence the possibility of an alternative name like `SparseConnectivityTracerProblems` or `SparseConnectivityTracerInstances`.

gdalle:
> Is that a big issue? I think removing some complexity at the cost of one dependency is acceptable.

This is pretty much a constant-complexity PR. It's just replacing the `include` with a `using`

> Wouldn't benchmark be a URL that's just as viable?

It's not just about giving a URL, it's about giving the URL of a package (as opposed to including a file)

test/nlpmodels.jl
```diff
@@ -33,7 +36,7 @@ Please look at the warnings displayed at the end.
 jac_inconsistencies = []
 hess_inconsistencies = []
 
-@testset "$name" for name in problem_names()
+@testset "$name" for name in Symbol.(OptimizationProblems.meta[!, :name])
```
adrhill, line 39: Maybe `problem_names` could be exported from `SparseConnectivityTracerBenchmarks` as well?

test/Project.toml
```diff
@@ -15,8 +16,10 @@ NLPModelsJuMP = "792afdf1-32c1-5681-94e0-d7bf7a5df49e"
 NNlib = "872c559c-99b0-510c-b3b7-b6c96a88d5cd"
 OptimizationProblems = "5049e819-d29b-5fba-b941-0eee7e64c1c6"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
+PkgJogger = "10150987-6cc1-4b76-abee-b1c1cbd91c01"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 ReferenceTests = "324d217c-45ce-50fc-942e-d289b448e8cf"
+SimpleDiffEq = "05bca326-078c-5bf0-a5bf-ce7c7982d7fd"
```
adrhill, line 22: Why did this have to be added?

benchmark/benchmarks.jl
```diff
@@ -1,30 +1,4 @@
...
-    SUITE["Jacobian"]["Global"][nameof(S1)] = jacbench(TracerSparsityDetector(G, H))
-    SUITE["Jacobian"]["Local"][nameof(S1)] = jacbench(TracerLocalSparsityDetector(G, H))
-    SUITE["Hessian"]["Global"][(nameof(S1), nameof(S2))] = hessbench(
-        TracerSparsityDetector(G, H)
-    )
-    SUITE["Hessian"]["Local"][(nameof(S1), nameof(S2))] = hessbench(
-        TracerLocalSparsityDetector(G, H)
-    )
-end
+@jog SparseConnectivityTracer
```
adrhill, line 3: This confused me at first and I had to refer to the PkgJogger docs.
Could you add a comment, e.g. *"Use `PkgJogger.@jog` to create the JogSparseConnectivityTracer package"*?

benchmark/SparseConnectivityTracerBenchmarks/src/nlpmodels.jl
```diff
@@ -1,18 +1,3 @@
...
-using ADNLPModels: ADNLPModels
-using NLPModels: NLPModels, AbstractNLPModel
-using NLPModelsJuMP: NLPModelsJuMP
-using OptimizationProblems: OptimizationProblems
-
-using Dates: now
-using LinearAlgebra
-using SparseArrays
-
-problem_names() = Symbol.(OptimizationProblems.meta[!, :name])
```
adrhill, line 14: Can this be kept and exported?

adrhill (changes requested):

test/Project.toml
gdalle: because it's used in the benchmarks

gdalle:
@adrhill the benchmarks work on this PR branch and fail on main but that's expected, this is good to merge

Issue #137, opened 2024-06-25
Store patterns in sparse vectors/matrices instead of sets


gdalle:
I want to try this but it might require larger code changes than I thought.
A first hurdle is `myempty`: for a set it doesn't require further information, but for a sparse array we need to give the ambient size.

adrhill:
I'll do this once #135 is merged.

adrhill:
Ah sorry, I just read your comment in #135.
Let me retry to add pattern types first, then we go back to this and #135.

adrhill:
Do you still want this or can I close the issue @gdalle? 

I can't imagine it being faster than `DictHessianPattern` since `SparseVector` and `SparseMatrixCSC` are based on `Vector`s of indices and values.

gdalle:
I think for small and medium problems using a BitMatrix to store the sparsity pattern in shared (Walther) hessian mode would crush the competition

gdalle:
So yeah I'd like to keep this issue open at the least

PR #139, opened 2024-06-26
Introduce sparsity patterns

adrhill:
This is the second attempt at #114 after #119 got derailed by unrelated compile time issues.

___

This PR introduces `AbstractPattern`s that represent sparsity patterns:
```
AbstractPattern
├── AbstractVectorPattern: used in GradientTracer, ConnectivityTracer
│   └── IndexSetVectorPattern
└── AbstractHessianPattern: used in HessianTracer
    └── IndexSetHessianPattern
```

Over the last dozens of PRs, we've kept refactoring the Tracer type signatures to accommodate new representations for sparse vectors and sparse matrices, most notably:
* moving away from shared first- and second-order information that were dictionaries of sets (#57)
* introducing custom data types, which had to be forced into the `AbstractSet` interface (#50)
* making second-order information more generic (#103)
* potentially adding traits for shared tracers (#135)

Advantages of introducing pattern types:
* It makes prototyping new ideas like #137 much faster, since tracer types don't have to be touched.
* It allows for the reintroduction of `DictHessianPattern`s, which can't be split into `gradient` and `hessian` fields.  
* PR #135 showed that dispatching on generic field types such as `AbstractSet` isn't enough, since `AbstractSet`s can be used in an allocating or non-allocating manner. 
In 135, the proposed workaround was to introduce a new field to the `HessianTracer` struct. This is problematic for several reasons:
  * It is harder to document than pattern structs, which have individual docstrings
  * If we support new field types, like #137, it implies that we always need to implement and maintain two variants, one that allocates and one that mutates
  * If the point above is circumvented, e.g. by using strict inner constructors, this once again puts burden on the documentation and/or users.

Recommended reading order for reviews:
1. `patterns.jl`
1. `tracers.jl` 
1. updated overload files

### DIFF
--- a/src/patterns.jl
+++ b/src/patterns.jl
@@ -6,13 +6,32 @@ Abstract supertype of all sparsity pattern representations.
 ## Type hierarchy
 ```
 AbstractPattern
-├── AbstractVectorPattern: used in GradientTracer, ConnectivityTracer
-│   └── IndexSetVectorPattern
+├── AbstractGradientPattern: used in GradientTracer, ConnectivityTracer
+│   └── IndexSetGradientPattern
 └── AbstractHessianPattern: used in HessianTracer
     └── IndexSetHessianPattern
 ```
 """
 abstract type AbstractPattern end
+
+"""
+  myempty(T)
+  myempty(tracer)
+  myempty(pattern)
+
+
+Constructor for an empty tracer or pattern of type `T` representing a new number (usually an empty pattern).
+"""
+myempty
+
+"""
+  seed(T, i)
+  seed(tracer, i)
+  seed(pattern, i)
+
+Constructor for a tracer or pattern of type `T` that only contains the given index `i`.
+"""
+seed
 
 #==========================#
 # Utilities on AbstractSet #
@@ -37,27 +56,26 @@ function union_product!(
 end
 
 #=======================#
-# AbstractVectorPattern #
+# AbstractGradientPattern #
 #=======================#
 
-# For use with ConnectivityTracer and GradientTracer.
+# For use with GradientTracer.
 
 """
-    AbstractVectorPattern <: AbstractPattern
+    AbstractGradientPattern <: AbstractPattern
 
 Abstract supertype of sparsity patterns representing a vector.
-For use with [`ConnectivityTracer`](@ref) and [`GradientTracer`](@ref).
+For use with [`GradientTracer`](@ref).
 
 ## Expected interface
 
 * `myempty(::Type{MyPattern})`: return a pattern representing a new number (usually an empty pattern)
 * `seed(::Type{MyPattern}, i::Integer)`: return an pattern that only contains the given index `i`
-* `inputs(p::MyPattern)`: return non-zero indices `i` for use with `ConnectivityTracer`
 * `gradient(p::MyPattern)`: return non-zero indices `i` for use with `GradientTracer`
 
 Note that besides their names, the last two functions are usually identical.
 """
-abstract type AbstractVectorPattern <: AbstractPattern end
+abstract type AbstractGradientPattern <: AbstractPattern end
 
 """
 $(TYPEDEF)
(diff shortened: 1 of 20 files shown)


adrhill:
Benchmarks show no regression and the only failing test is JET:
image

I'm not sure what the issue is.

src/patterns.jl
```diff
@@ -0,0 +1,131 @@
...
+    AbstractPattern
+
+Abstract supertype of all sparsity pattern representations.
+
+## Type hierarchy
+```
+AbstractPattern
+├── AbstractVectorPattern: used in GradientTracer, ConnectivityTracer
+│   └── IndexSetVectorPattern
+└── AbstractHessianPattern: used in HessianTracer
```
gdalle, line 11: Why not `AbstractMatrixVectorPattern`? The naming would be more coherent

src/patterns.jl
```diff
@@ -0,0 +1,131 @@
...
+"""
+    AbstractVectorPattern <: AbstractPattern
+
+Abstract supertype of sparsity patterns representing a vector.
+For use with [`ConnectivityTracer`](@ref) and [`GradientTracer`](@ref).
+
+## Expected interface
+
+* `myempty(::Type{MyPattern})`: return a pattern representing a new number (usually an empty pattern)
+* `seed(::Type{MyPattern}, i::Integer)`: return an pattern that only contains the given index `i`
```
gdalle, line 54: These won't work for array-backed patterns because of #137: we need the actual length of the input at some point (perhaps even the size). Input size information should only be stored in the pattern object itself, not in its type (unless we want a compilation nightmare).

src/patterns.jl
```diff
@@ -0,0 +1,131 @@
+"""
+    AbstractPattern
+
+Abstract supertype of all sparsity pattern representations.
+
+## Type hierarchy
+```
+AbstractPattern
+├── AbstractVectorPattern: used in GradientTracer, ConnectivityTracer
+│   └── IndexSetVectorPattern
```
gdalle, line 10: I think it would be a good idea to flesh out an array-backed pattern in this same PR, to anticipate potential problems like carrying the size (see below)

src/patterns.jl
```diff
@@ -0,0 +1,131 @@
...
+
+Abstract supertype of sparsity patterns representing a vector.
+For use with [`ConnectivityTracer`](@ref) and [`GradientTracer`](@ref).
+
+## Expected interface
+
+* `myempty(::Type{MyPattern})`: return a pattern representing a new number (usually an empty pattern)
+* `seed(::Type{MyPattern}, i::Integer)`: return an pattern that only contains the given index `i`
+* `inputs(p::MyPattern)`: return non-zero indices `i` for use with `ConnectivityTracer`
+* `gradient(p::MyPattern)`: return non-zero indices `i` for use with `GradientTracer`
```
gdalle, line 56: Why differentiate between these?

src/patterns.jl
```diff
@@ -0,0 +1,131 @@
...
+
+"""
+$(TYPEDEF)
+
+Vector sparsity pattern represented by an `AbstractSet` of indices ``{i}`` of non-zero values.
+
+## Fields
+$(TYPEDFIELDS)
+"""
+struct IndexSetVectorPattern{I<:Integer,S<:AbstractSet{I}} <: AbstractVectorPattern
```
gdalle, line 70: I think we can use this PR to remove `AbstractSet` subtyping (because again, our custom types don't satisfy every assumption) and use duck-typing for that stuff

src/tracers.jl
```diff
@@ -199,27 +184,46 @@ end
 # Utilities #
 #===========#
 
-myempty(::Type{ConnectivityTracer{I}}) where {I} = ConnectivityTracer{I}(myempty(I), true)
-myempty(::Type{GradientTracer{G}}) where {G}     = GradientTracer{G}(myempty(G), true)
-myempty(::Type{HessianTracer{G,H}}) where {G,H}  = HessianTracer{G,H}(myempty(G), myempty(H), true)
+"""
+  myempty(T)
+  myempty(tracer)
+  myempty(pattern)
```
gdalle, line 190: shouldn't this go in the `patterns.jl` file?

gdalle (changes requested):

src/patterns.jl
adrhill: Ideally, I would have called `AbstractVectorPattern` `AbstractGradientPattern`, but there is the permanently annoying `ConnectivityTracer`.

Do you prefer `AbstractFirstOrderPattern` and `AbstractSecondOrderPattern`?
I'm just a bit afraid of an explosion in subtype-name-length, e.g. `IndexSetSecondOrderPattern`.

Maybe there is a shorter alternative to `FirstOrder` and `SecondOrder`?

src/patterns.jl
adrhill: > Input size information should only be stored in the pattern object itself, not in its type (unless we want a compilation nightmare)

I honestly don't think it would be too bad for realistic use-cases. The input size is constant over the entire sparsity tracing procedure, and therefore only one type gets created.

src/patterns.jl
adrhill: Because the pattern represents something different when used in a `ConnectivityTracer` or `GradientTracer`. I could just call it `vector`, but I feel like that's also an odd name for `ConnectivityTracer`.

Suggestions are welcome, I fully agree that a single function would be best!

src/patterns.jl
adrhill: Otherwise, if needed, we can still modify this to `seed(T, i, n)`.

src/patterns.jl
adrhill: I'd rather do this in another PR since it requires input enumeration.
I have some ideas for this I hinted at in #28. I'd also try to support sparse-matrix inputs, the need for which popped up in #131.

src/patterns.jl
adrhill: My idea is that stuff like `DuplicateVector` gets to be a `DuplicatePattern` in the future.

gdalle: Okay that's fair. But then we should temporarily remove it from the API if we're gonna change its place in the hierarchy

src/patterns.jl
gdalle: That's because you only think of one application of the tracing. What if we need to trace local sparsity on each array of a dataset, each having a different size (e.g. for a dataset of graphs)?

src/patterns.jl
adrhill: You mean before tagging a release?

adrhill:
Let's drop `ConnectivityTracer` in a follow-up PR. I will rename new features in here accordingly. That should make it easier to revert the removal in case anyone ever wants `ConnectivityTracer` back. 

I will merge this PR without input size information to avoid feature creep. However, I will refrain from tagging a release without it.

src/patterns.jl
adrhill: Will be addressed by removing `ConnectivityTracer` in a follow-up PR.

src/patterns.jl
adrhill: Let's do this in a follow up. It's a not a problem introduced by this PR, but one that also exists on main. We'll deal with it before tagging a breaking release.

src/patterns.jl
adrhill: I renamed it to `AbstractGradientPattern` in anticipation of removing `ConnectivityTracer` in a follow-up PR.

src/patterns.jl
gdalle: Yeah, let's not tag a release where DuplicateVector is an abstract set and then change it right away to a pattern

PR #140, opened 2024-06-27
Remove `ConnectivityTracer` and legacy API

adrhill:
This PR removes `ConnectivityTracer`. 
Maybe this is also a good time to completely abandon the old native API?


gdalle:
> Maybe this is also a good time to completely abandon the old native API?

Sounds good, let's use ADTypes in the tests too

adrhill:
> Sounds good, let's use ADTypes in the tests too

There isn't going to be any other way to pass the tests. ;)

I've opted for re-exporting `jacobian_sparsity` and `hessian_sparsity`. Do you have strong opinions on this?

PR #146, opened 2024-08-06
Support ComponentArrays

gdalle:
- In `trace_input`, ensure that the array of indices `is` has the same type as the input array `xs`
- Add tests for ComponentArrays


src/interface.jl
```diff
@@ -17,7 +17,8 @@ Supports [`GradientTracer`](@ref), [`HessianTracer`](@ref) and [`Dual`](@ref).
 trace_input(::Type{T}, xs) where {T<:Union{AbstractTracer,Dual}} = trace_input(T, xs, 1)
 
 function trace_input(::Type{T}, xs::AbstractArray, i) where {T<:Union{AbstractTracer,Dual}}
-    is = reshape(1:length(xs), size(xs)) .+ (i - 1)
+    is = similar(xs, Int)  # same array type as xs
+    is .= reshape(1:length(xs), size(xs)) .+ (i - 1)
```
adrhill, line 21: It looks like this in-place assignment fails on some matrix types like `Diagonal` and `Symmetric`.

adrhill (changes requested):
Looks good, thanks!

adrhill:
I've opened a related issue in #147 describing the problem at hand for `Diagonal`.

We are going to have to add more `trace_input` methods for common matrix types.

gdalle:
In the meantime I would rather merge (some version of) this, which is more correct, and possibly remove tests on `Symmetric` and `Diagonal`, than keep the current broken behavior?

adrhill:
Absolutely don't remove tests!

I could live with you copying the previous implementation of `trace_input` as an additional method on `Diagonal`, `Symmetric` and Co, making the new one the default for `AbstractArray`.

adrhill:
And then we fix those new methods and close #147 in a follow-up PR.

gdalle:
The tests we have are a bit inadequate to say the least, because they didn't catch this reshape bug. And this bug in turn underlines that we don't even know what the right semantics are for tracing from a structured array 🤷‍♂️

adrhill:
We currently just are too conservative with the sparsity pattern on `Diagonal`.

The way I see it, support for ComponentArrays is orthogonal to making patterns on  `Diagonal` less conservative.

If you remove the tests and support for `Diagonal` entirely, that would put this repo in a temporarily breaking state for no reason. Users would run into the same error the tests run into. Avoiding this just requires 5 additional LOC which already exist.

gdalle:
Not sure what fix you have in mind, can I let you show me?

gdalle:
Perhaps we should just call parent before we do anything

adrhill (approved):

PR #151, opened 2024-08-07
Increase performance via symmetry of Hessian

### LINKED ISSUE #34: Potential increase in performance via symmetry of Hessian
The Hessian tracer added in #20 keeps track of all first and second-order interactions between indices. However, these aren't all needed due to symmetry.
To make an analogy using Polynomials, we separately keep track of both $x_1x_2$ and $x_2x_1$.

By adding numerical comparisons between indices before updating sets, we could potentially speed up performance on some set types outside of `BitSet`s. 

It would be good to add a benchmarking suite with multiple set types before working on this.

adrhill:
Closes #34.

By only keeping track of index-tuples $(i, j)$ with $i≤j$, the size of all "Hessian-sets" is reduced by roughly half. 
As seen in #135, the size of sets plays a large role in the performance of SCT as it affects the performance of union operations.

~~This PR changes the output type of `hessian_sparsity` from `SparseMatrixCSC{Bool, Int64}` to `Symmetric{Bool, SparseArrays.SparseMatrixCSC{Bool, Int64}}`.~~
Is this (a) desired and (b) should this be considered breaking? (Ping @gdalle for the matrix coloring perspective)

### DIFF
--- a/src/adtypes.jl
+++ b/src/adtypes.jl
@@ -24,7 +24,7 @@ julia> using ADTypes, SparseConnectivityTracer
 julia> f(x) = x[1] + x[2]*x[3] + 1/x[4];
 
 julia> hessian_sparsity(f, rand(4), TracerSparsityDetector())
-4×4 SparseArrays.SparseMatrixCSC{Bool, Int64} with 3 stored entries:
+4×4 LinearAlgebra.Symmetric{Bool, SparseArrays.SparseMatrixCSC{Bool, Int64}}:
  ⋅  ⋅  ⋅  ⋅
  ⋅  ⋅  1  ⋅
  ⋅  1  ⋅  ⋅
@@ -91,7 +91,7 @@ julia> using ADTypes, SparseConnectivityTracer
 julia> f(x) = x[1] + max(x[2], x[3]) * x[3] + 1/x[4];
 
 julia> hessian_sparsity(f, [1.0, 2.0, 3.0, 4.0], TracerLocalSparsityDetector())
-4×4 SparseArrays.SparseMatrixCSC{Bool, Int64} with 2 stored entries:
+4×4 LinearAlgebra.Symmetric{Bool, SparseArrays.SparseMatrixCSC{Bool, Int64}}:
  ⋅  ⋅  ⋅  ⋅
  ⋅  ⋅  ⋅  ⋅
  ⋅  ⋅  1  ⋅
--- a/src/adtypes.jl
+++ b/src/adtypes.jl
@@ -24,7 +24,7 @@ julia> using ADTypes, SparseConnectivityTracer
 julia> f(x) = x[1] + x[2]*x[3] + 1/x[4];
 
 julia> hessian_sparsity(f, rand(4), TracerSparsityDetector())
-4×4 LinearAlgebra.Symmetric{Bool, SparseArrays.SparseMatrixCSC{Bool, Int64}}:
+4×4 SparseArrays.SparseMatrixCSC{Bool, Int64} with 3 stored entries:
  ⋅  ⋅  ⋅  ⋅
  ⋅  ⋅  1  ⋅
  ⋅  1  ⋅  ⋅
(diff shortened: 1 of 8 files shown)


adrhill:
> I would say this is breaking, yes. 

Should we make the change with our next breaking release though?

> Maybe we could just do sparse(Symmetric(sparse(...))) 

I've handled it manually. :)

adrhill:
I've never seen a single Hessian result appear in any of the hundreds of Benchmark results we've produced over the years. Maybe they get discarded because they get run with too few samples? I can't believe this PR didn't result in a performance improvement.

gdalle:
If you unfold and `Ctrl+F` Hessian, you can see that the Hessian benchmarks run on the "target" code (aka the `main` branch). What's weird is they don't run on the "baseline" code (aka the current PR).

gdalle:
Even weirder, when you check the CI logs, the Hessian benchmarks also run for the baseline branch

adrhill:
I get a decent speed-up when running the `arrowhead` benchmark locally:

### This PR
```Julia-repl
julia> include("benchmark/hessian.jl")
hessbench_randomsparsity (generic function with 1 method)

julia> method = TracerSparsityDetector()
TracerSparsityDetector{SparseConnectivityTracer.GradientTracer{SparseConnectivityTracer.IndexSetGradientPattern{Int64, BitSet}}, SparseConnectivityTracer.HessianTracer{SparseConnectivityTracer.IndexSetHessianPattern{Int64, BitSet, Set{Tuple{Int64, Int64}}, false}}}()

julia> N, K = 200, 16
(200, 16)

julia> x = rand(N);

julia> f = ArrowHead(K)
ArrowHead(16)

julia> @benchmark hessian_sparsity($f, $x, $method)
BenchmarkTools.Trial: 389 samples with 1 evaluation.
 Range (min … max):  11.577 ms …  16.529 ms  ┊ GC (min … max): 0.00% … 23.09%
 Time  (median):     12.855 ms               ┊ GC (median):    8.48%
 Time  (mean ± σ):   12.845 ms ± 548.286 μs  ┊ GC (mean ± σ):  7.83% ±  3.41%

                           ▆▆▆▆▅▆█▁▂▁▃▂                         
  ▂▄▄▆▄▇▄▄▃▃▃▁▃▂▁▁▁▂▁▂▄▄▆▄▆████████████▇▄▆▄▂▃▄▃▃▂▄▃▃▂▂▄▂▁▄▂▂▃▄ ▄
  11.6 ms         Histogram: frequency by time         14.1 ms <

 Memory estimate: 42.39 MiB, allocs estimate: 98757.
```

### `main` branch
```Julia-repl
julia> @benchmark hessian_sparsity($f, $x, $method)
BenchmarkTools.Trial: 222 samples with 1 evaluation.
 Range (min … max):  20.985 ms …  27.925 ms  ┊ GC (min … max): 5.15% … 22.89%
 Time  (median):     22.470 ms               ┊ GC (median):    6.71%
 Time  (mean ± σ):   22.575 ms ± 899.500 μs  ┊ GC (mean ± σ):  7.99% ±  2.95%

         ▄▄    █  ▄▃▅  ▂▅▃ ▃ ▂▃▃ ▅▂                             
  ▃▁▃▃▅▆▆██▇██▇██████▅████▇█▇███▇███▅██▅▅▅▇▃▅▃▃▅▇▁▁▅▁▃▁▃▃▁▃▃▁▅ ▅
  21 ms           Histogram: frequency by time         24.9 ms <

 Memory estimate: 67.95 MiB, allocs estimate: 106189.

adrhill:
> If you unfold and Ctrl+F Hessian, you can see that the Hessian benchmarks run on the "target" code (aka the main branch). What's weird is they don't run on the "baseline" code (aka the current PR).

> Even weirder, when you check the CI logs, the Hessian benchmarks also run for the baseline branch

It looks like the report summarizes the baseline as well, but for some reason doesn't put it in the summary table. A speedup is clearly visible:

| PR  | main |
|:---:|:---:|
| Screenshot 2024-08-08 at 15 30 03 | Screenshot 2024-08-08 at 15 30 16 |

Issue #147, opened 2024-08-06
Add custom `trace_input` methods for LinearAlgebra matrix types


adrhill:
As shown in #146, if the user inputs a 3x3 `Diagonal` matrix, we currently create a dense `Matrix` of tracers with input indices 1 to 9.
This leads to non-conservative sparsity patterns:

```Julia
julia> x = Diagonal(rand(3))
3×3 Diagonal{Float64, Vector{Float64}}:
 0.674044   ⋅         ⋅ 
  ⋅        0.683069   ⋅ 
  ⋅         ⋅        0.387561

julia> jacobian_sparsity(sum, x, TracerSparsityDetector())
1×9 SparseArrays.SparseMatrixCSC{Bool, Int64} with 9 stored entries:
 1  1  1  1  1  1  1  1  1
```

Even worse, the user-provided function is called with a `Matrix` of tracers instead of a `Diagonal` of tracers, likely hitting an entirely different method.

For the example above, a custom `trace_input` method should be implemented that returns a `Diagonal` matrix of three tracers with indices 1, 5 and 9.

adrhill:
While `Diagonal` is easily fixed, `Symmetric` is much more tricky: 
For a 3x3 `Symmetric` matrix, we actually want to create 9 tracers with input indices 1 to 9. However, the returned matrix can't be symmetric!

Issue #145, opened 2024-08-06
Add support for `ComponentVector`


SouthEndMusic:
Currently tracing through a function with `ComponentVector` input and output does not work. E.g.

```julia
using SparseConnectivityTracer: TracerSparsityDetector, jacobian_sparsity
using ComponentArrays: ComponentVector
detector = TracerSparsityDetector()

function f!(output, input)
    output.x .= input.y
    output.y .= input.x
end

input = ComponentVector{Float64}(; x=rand(5), y=rand(5))
output = zero(input)

jacobian_sparsity(f!, output, input, detector)
```

yields

```
type Array has no field y
```

gdalle:
Can you try out the branch from #146 on your real use case, see if other problems pop up?

SouthEndMusic:
I gave it a try, and now it breaks down on `DataInterpolations`, specifically https://github.com/SciML/DataInterpolations.jl/blob/35db42ea22fae1caeb92ca6cfbb4a204ace4e401/src/interpolation_methods.jl#L2. You document that you do not support `<, >`, but I feel like this should work as it is a 'dead end' of the trace that can be ignored

gdalle:
What do you mean by dead end that can be ignored?

gdalle:
Note that SparseConnectivityTracer errors at the comparison. It has no way to know that an error is thrown right after

SouthEndMusic:
The jacobian sparsity from symbolics doesn't error on this, so I wonder what is going on there.

I can imagine that the tracing goes wrong if the path trough the code branches out based on the input in a way that affects the output. But that is not what happens here

gdalle:
Yeah the Symbolics approach works in a different way, and can more easily handle conditionals than our operator overloading method.
The reason it fails is that `TracerSparsityDetector` aims to be independent of the input value, which is why we can't return anything meaningful for `x < y` (we don't know if it's `true` or `false`).
As a workaround, if you're 100% sure that the sparsity pattern does not depend on the input value, you can use `TracerLocalSparsityDetector`. But we will probably have to add custom overloads for DataInterpolations anyway, as noted in https://github.com/adrhill/SparseConnectivityTracer.jl/issues/74#issuecomment-2256399578

SouthEndMusic:
With `TracerLocalSparsityDetector` it fails with

```
[...]

PR #165, opened 2024-08-13
Increase sparsity for `Diagonal` inputs

adrhill:
By having `trace_input` create a `Diagonal` matrix of tracers with only the indices from `diagind`, the sparsity of patterns can be increased by a lot on `Diagonal` inputs:

image

Previously, this would have returned a dense pattern of ones (see #147).

### DIFF
--- a/test/test_arrays.jl
+++ b/test/test_arrays.jl
@@ -47,15 +47,19 @@ pow3(A) = A^3
 
 method = TracerSparsityDetector()
 
-test_all_one(A) = @test all(isone, A)
-test_all_zero(A) = @test all(iszero, A)
+allone(A) = all(isone, A)
+allzero(A) = all(iszero, A)
 
 # Short-hand for Jacobian pattern of `x -> sum(f(A))`
 Jsum(f, A) = jacobian_sparsity(SumOutputs(f), A, method)
 # Test whether all entries in Jacobian are zero
-testJ0(f, A) = @testset "Jacobian" test_all_zero(Jsum(f, A))
+testJ0(f, A) = @testset "Jacobian" begin
+    @test allzero(Jsum(f, A))
+end
 # Test whether all entries in Jacobian are one where inputs were non-zero.
-testJ1(f, A) = @testset "Jacobian" test_all_one(Jsum(f, A))
+testJ1(f, A) = @testset "Jacobian" begin
+    @test allone(Jsum(f, A))
+end
 function testJ1(f, A::Diagonal)
     @testset "Jacobian" begin
         jac = Jsum(f, A)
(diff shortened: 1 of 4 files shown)


test/test_arrays.jl
```diff
@@ -98,56 +240,102 @@ end
...
+        ]
+    end
+    @testset "pow0 $(arrayname(A))" for A in NONDIAG_SQUARE_MATRICES
+        testJ0(pow0, A)
+        testH0(pow0, A)
+    end
+    @testset "pow0 $(arrayname(A))" for A in DIAG_SQUARE_MATRICES
+        # TODO: these should be zero and are currently too conservative
+        @test_broken all(iszero, Jsum(pow0, A))
+        @test_broken all(iszero, Hsum(pow0, A))
```
adrhill, line 285: To return empty Jacobians and Hessians on `pow(A, 0)`, we will need some additional methods on `Diagonal` matrices of tracers.

adrhill: Fixed in [bb16995](https://github.com/adrhill/SparseConnectivityTracer.jl/pull/165/commits/bb169953622662c8fdb9550c72c41f09e6fe3594)

gdalle:
Is this even necessary?
I understand the need to handle `Diagonal` matrices _created inside_ the function we trace. But I strongly doubt that there is a practical use case in which people would pass a `Diagonal` as _input_ of sparsity detection?

adrhill:
This issue started with #145, where a user had a `ComponentArray` input. 
Until then, every input was converted to a dense `Matrix`.

gdalle:
I think this PR does two things: 1) more overloads for Diagonal inside generic code and 2) handling a Diagonal as input of sparsity detection. I'm only questioning the legitimacy of 2. While ComponentArrays or Symmetric matrices as input to sparsity detection make sense, Diagonal matrices make less sense to me.

adrhill:
2) is 4 LOC and easier testing of 1) already makes it worth it, even if no user ever has `Diagonal` inputs.

gdalle:
I'm just asking where you draw the line. To push this reasoning to the extreme, we want code to work if it uses FillArrays internally, but no sane person would give a FillArray as input of sparsity detection

adrhill:
> I'm just asking where you draw the line.

I definitely want to support `SparseArray` inputs in the future.

As soon as we "internally" support an array type, it makes sense to write a matching `trace_input` method. It makes testing and debugging that much easier.

> To push this reasoning to the extreme, we want code to work if it uses FillArrays internally, but no sane person would give a FillArray as input of sparsity detection

Your point reminds me that we don't explicitly test FillArrays yet.

Issue #196, opened 2024-10-01
Interoperability with dual numbers 


SouthEndMusic:
I want to obtain the Jacobian sparsity of a function whose body contains a `ForwardDiff.derivative` call. This leads to this error:

```
ERROR: MethodError: *(::SparseConnectivityTracer.GradientTracer{...}, ::ForwardDiff.Dual{...}) is ambiguous.

Possible fix, define
  *(::SparseConnectivityTracer.GradientTracer, ::ForwardDiff.Dual{Ty}) where Ty
```

Does this have a quick fix?

_Edit: I worked around it, so you can close the issue if you don't find this important._

adrhill:
Maybe we're running into the issue described here: https://docs.julialang.org/en/v1/manual/methods/#Abstract-containers-and-element-types

adrhill:
If this is a desired feature, we should rework the function overloading utilities for 2-to-1 functions to accept an optional type argument.

Then you could add a package extension for ForwardDiff and call something like
```Julia
eval(SCT.generate_code_2_to_1(:Base, Dual, ops_1_to_1))
```

However, I'm not sure we can make this work in combination with other package extensions.

gdalle:
Essentially the more general takeaway is that SCT doesn't play nice with other packages based on operator overloading

gdalle:
Maybe for operations between a `Tracer` and a number we can handle things better through [promotion](https://docs.julialang.org/en/v1/manual/conversion-and-promotion/#Promotion)?

adrhill:
> Essentially the more general takeaway is that SCT doesn't play nice with other packages based on operator overloading

More generally packages implementing 2-to-1 Base functions.

> the solution was promotion. But the error wasn't a method error so idk.

I'm not sure how we would hit type promotion if the method call itself is ambiguous. 
I'll open a draft PR and see how much complexity it adds.

adrhill:
Basic operators from Julia Base should now work with `ForwardDiff.Dual` on main.
I'll tag a release tomorrow.

adrhill:
After thinking about it for a day, I'm starting to think #200 shouldn't be released.
With #200, `ForwardDiff.Dual` are treated just like any other `Real` number, meaning that multiplication with a global tracer basically ignores the `ForwardDiff.Dual`. However, `ForwardDiff.Dual` could in theory contain SCT's tracers, thus returning wrong, non-conservative sparsity patterns.

[...]

PR #197, opened 2024-10-01
Generate 2-to-1 overloads on arbitrary types

adrhill:
Previously, 2-to-1 overloads were hardcoded on `Real` numbers.
This PR generalizes this to arbitrary types, simplifying the code needed to avoid ambiguity errors. 
This is also the first step towards a ForwardDiff extension addressing #196.


adrhill:
JET.jl and DataInterpolations tests fail, but it looks unrelated to this PR.

PR #201, opened 2024-10-01
Fix detector display by replacing `println` with `print`

gdalle:
This fixes the following display:

```julia
julia> using ADTypes, SparseConnectivityTracer

julia> backend = AutoSparse(AutoForwardDiff(), sparsity_detector=TracerSparsityDetector())
AutoSparse(dense_ad=AutoForwardDiff(), sparsity_detector=TracerSparsityDetector()
)
```


adrhill:
Thanks!

PR #211, opened 2024-11-05
Change default set type to `Set{Int}`

adrhill:


adrhill:
Closing in favor of #218.

PR #232, opened 2025-03-11
Add stable API for allocation of buffers

adrhill:
For use in external packages like DI (see https://github.com/JuliaDiff/DifferentiationInterface.jl/issues/737).

I'm open to a name change for the two functions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # SparseConnectivityTracer.jl
 
 ## Version `v0.6.14`
-* ![Feature][badge-feature] Add stable API for allocation of buffers via `jacobian_buffer` and `hessian_buffer`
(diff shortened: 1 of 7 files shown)


gdalle (approved):
lgtm!

SouthEndMusic:
This does not yet work for my use case, because the Jacobian buffer gets initialized with `undef` which yields an `UndefRefError`. In my code it can happen that I read from the buffer before it is written to. This serves no particular function and I could work around it, but it would be nice if that wasn't necessary.

gdalle:
> In my code it can happen that I read from the buffer before it is written to. This serves no particular function and I could work around it, but it would be nice if that wasn't necessary.

I don't understand how this can happen without being a mistake, could you give an example?

SouthEndMusic:
In my code a certain `du` entry can be set either based on a static parameter value or based on a state-derived value in the cache, depending on certain discrete model states. I access both values, but use only one. I could refactor so that I only acces the state-derived value in the cache if that is the right data source

PR #233, opened 2025-03-18
Add stable API for tracer type

gdalle:
This is needed in DI for caches that are more complex than arrays: I need access to the element type and not just to a `similar` variant because I write my own recursive `similar` variant.

No need for additional testing since I use the new functions functions inside `jacobian_buffer` and `hessian_buffer`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # SparseConnectivityTracer.jl
 
+## Version `v0.6.15`
+* ![Feature][badge-feature] Add stable API for tracer type via `jacobian_tracer_type` and `hessian_tracer_type` ([#233])
+
 ## Version `v0.6.14`
 * ![Feature][badge-feature] Add stable API for allocation of buffers via `jacobian_buffer` and `hessian_buffer` ([#232])
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # SparseConnectivityTracer.jl
 
 ## Version `v0.6.15`
-* ![Feature][badge-feature] Add stable API for tracer type via `jacobian_tracer_type` and `hessian_tracer_type` ([#233])
+* ![Feature][badge-feature] Add stable API for tracer type via `jacobian_eltype` and `hessian_eltype` ([#233])
 
 ## Version `v0.6.14`
 * ![Feature][badge-feature] Add stable API for allocation of buffers via `jacobian_buffer` and `hessian_buffer` ([#232])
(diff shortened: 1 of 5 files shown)


src/adtypes_interface.jl
```diff
@@ -172,18 +172,35 @@ for detector in (:TracerSparsityDetector, :TracerLocalSparsityDetector)
...
+"""
+    hessian_tracer_type(x, detector)
+
+Return the tracer type used for Hessian sparsity detection.
+"""
+hessian_tracer_type(x, ::TracerSparsityDetector{TG,TH}) where {TG,TH} = TH
+function hessian_tracer_type(x, ::TracerLocalSparsityDetector{TG,TH}) where {TG,TH}
+    return Dual{eltype(x),TH}
+end
+
```
adrhill, line 193: Adding `x::Real` might make sense here.

Duals expect `Dual{P<:Real,T<:AbstractTracer}` and we probably also don't want to return a global tracer when passed an array.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # SparseConnectivityTracer.jl
 
+## Version `v0.6.15`
+* ![Feature][badge-feature] Add stable API for tracer type via `jacobian_tracer_type` and `hessian_tracer_type` ([#233])
```
adrhill, line 4: URL for `[#233]` is missing at the bottom.

adrhill (changes requested):

src/adtypes_interface.jl
gdalle: Solved, assuming you meant `AbstractArray{<:Real}`

src/adtypes_interface.jl
```diff
@@ -172,18 +172,43 @@ for detector in (:TracerSparsityDetector, :TracerLocalSparsityDetector)
 end
 
 ## Stable API to allow packages like DI to allocate caches of tracers
+
+"""
+    jacobian_tracer_type(x, detector)
+
+Return the tracer type used for Jacobian sparsity detection.
+"""
+jacobian_tracer_type(x::AbstractArray{<:Real}, ::TracerSparsityDetector{TG}) where {TG} = TG
```
adrhill, line 181: If `x` is supposed to be an `AbstractArray`, we might want to call these `jacobian_tracer_eltype` and support scalars as well?

```Julia
julia> eltype(rand(Float32, 2))
Float32

julia> eltype(0.0f0)
Float32
```

Would be a bit more intuitive to use if it mirrored a known Julia function.

adrhill (changes requested):

adrhill (approved):

PR #238, opened 2025-04-03
Recursive array tools support

### LINKED ISSUE #237: Add support for `RecursiveArrayTools.jl`
Currently running sparsity detection with a `NamedArrayPartition` fails:

```julia
using RecursiveArrayTools: NamedArrayPartition

function f!(du, u)
    du.foo .= u.foo
end

u = NamedArrayPartition(; foo=rand(5))
du = copy(u)

jacobian_sparsity(f!, du, u, TracerSparsityDetector())
```

with an `Error: type Array has no field foo`.

SouthEndMusic:
Fixes https://github.com/adrhill/SparseConnectivityTracer.jl/issues/237.


SouthEndMusic:
@adrhill I implemented `trace_input` methods in an extension and these are indeed triggered in my test, but the error remains

No idea where that DataInterpolations stuff comes from btw

test/recursive_array_tools.jl
```diff
@@ -0,0 +1,12 @@
+using RecursiveArrayTools
```
adrhill, line 1: Looks like this is missing from the test environment.

test/recursive_array_tools.jl
```diff
@@ -0,0 +1,12 @@
...
+function f!(du, u)
+    du.foo[2] = u.foo[1]
+    du.foo[1] = u.foo[2]
+    return nothing
+end
+
+u = NamedArrayPartition(; foo=rand(5))
+du = copy(u)
+
+jacobian_sparsity(f!, du, u, TracerSparsityDetector())
```
adrhill, line 12: Would be nice to have a slightly more rigorous test than just a dry-run. 

I don't know the functionality of `NamedArrayPartition`, but the tests should cover as much of it as possible.

adrhill:
So `ArrayPartition` are always flat vectors? Does this support scalar values, e.g. `ArrayPartition(rand(2,2), 1.0, rand(3))`?

PR #239, opened 2025-04-15
Performance optimization in pattern creation

gdalle:
I stumbled upon a dynamic dispatch when profiling for https://github.com/JuliaDiff/DifferentiationInterface.jl/issues/774. Here's the simplest possible test case:

```julia
sparsity_detector = TracerSparsityDetector()
x, y = ones(10^6), ones(10^6);
jacobian_sparsity(copyto!, y, x, sparsity_detector)
```

And here's the flame graph:

Screenshot 2025-04-15 at 22 45 07

My fix removes the type instability and speeds up this test case by half:

```julia
julia> using Chairmarks

julia> @b jacobian_sparsity(copyto!, y, x, sparsity_detector) seconds=5  # before
119.358 ms (3999614 allocs: 268.377 MiB, 14.12% gc time)

julia> @b jacobian_sparsity(copyto!, y, x, sparsity_detector) seconds=5  # after
49.274 ms (3000114 allocs: 253.125 MiB, 24.52% gc time)
```

### DIFF
--- a/src/patterns.jl
+++ b/src/patterns.jl
@@ -180,7 +180,7 @@ function myempty(::Type{IndexSetGradientPattern{I,S}}) where {I,S}
     return IndexSetGradientPattern{I,S}(myempty(S))
 end
 function create_patterns(::Type{P}, xs, is) where {I,S,P<:IndexSetGradientPattern{I,S}}
-    sets = seed.(Ref(S), is)
+    sets = map(Base.Fix1(seed, S), is)
     return P.(sets)
 end
 


gdalle:
Edit: I just understood what went wrong with your broadcasting approach. You used `Ref` to encapsulate a fixed argument, as one does, except that this fixed argument was a type. Putting it into `Ref` prevented the compiler from specializing on it: all it remembered afterwards was that `S isa DataType`

```julia
julia> S = Set{Int}
Set{Int64}

julia> Ref(S)
Base.RefValue{DataType}(Set{Int64})

julia> typeof(Ref(S))
Base.RefValue{DataType}
```

adrhill:
TIL, nice! I'll tag a quick patch release to help with the ongoing benchmarks in https://github.com/JuliaDiff/DifferentiationInterface.jl/issues/774.

PR #241, opened 2025-04-15
Add matrix multiplication shortcuts

gdalle:
Modeled after matrix division.

Still need to do rowwise and columnwise reduction. Also we need better tests in case A and B share indices

### DIFF
--- a/src/overloads/arrays.jl
+++ b/src/overloads/arrays.jl
@@ -119,14 +119,14 @@ function Base.:*(A::Matrix{T}, B::Matrix) where {T<:AbstractTracer}
     if size(A, 2) != size(B, 1)
         throw(DimensionMismatch("arguments must have compatible dimensions"))
     end
-    t = second_order_or(A)
+    t = first_order_or(A)
     return Fill(t, size(A, 1), size(B, 2))
 end
 function Base.:*(A::Matrix{T}, B::Vector) where {T<:AbstractTracer}
     if size(A, 2) != length(B)
         throw(DimensionMismatch("arguments must have compatible dimensions"))
     end
-    t = second_order_or(A)
+    t = first_order_or(A)
     return Fill(t, size(A, 1))
 end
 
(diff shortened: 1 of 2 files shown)


src/overloads/arrays.jl
```diff
@@ -114,6 +114,56 @@ for (Tx, TA, Ty) in Iterators.filter(
     @eval LinearAlgebra.dot(x::$Tx, A::$TA, y::$Ty) = LinearAlgebra.dot(x, A * y)
 end
 
+## Multiplication
+function Base.:*(A::Matrix{T}, B::Matrix) where {T<:AbstractTracer}
```
adrhill, line 118: We probably only want this on `GradientTracer`.

src/overloads/arrays.jl
```diff
@@ -114,6 +114,56 @@ for (Tx, TA, Ty) in Iterators.filter(
...
+    if size(A, 2) != size(B, 1)
+        throw(DimensionMismatch("arguments must have compatible dimensions"))
+    end
+    t = map(first_order_or, eachrow(A))
```
adrhill, line 122: Multiplication is twice differentiable in the cross-derivative, so this has to be `second-order-or`. 
There won't be a difference for `GradientTracer`, but there is no reason to use the wrong one here.

adrhill (changes requested):
Same two comments apply to all methods.

src/overloads/arrays.jl
gdalle: Why? I think this is also correct for Hessian tracing.

src/overloads/arrays.jl
gdalle: The line you commented performs the equivalent of the sum of each row. So I don't think we need second order,

src/overloads/arrays.jl
adrhill: Exactly. 👍

src/overloads/arrays.jl
gdalle: What I'm saying is: I think this is correct even for Hessian tracing

adrhill:
Learning from the DataInterpolations extension, we might also want to subtype all non-tracer matrices and vectors as `Matrix{<:Number}`.

PR #243, opened 2025-05-02
Stop tracing through multiplication by zero

### LINKED ISSUE #242: Take into account multiplication by fixed (non-dual) 0
```julia
julia> using SparseConnectivityTracer

julia> detector = TracerSparsityDetector()
TracerSparsityDetector()

julia> multzero(x) = [x[1] * 0]
multzero (generic function with 1 method)

julia> jacobian_sparsity(multzero, [1.0], detector)
1×1 SparseArrays.SparseMatrixCSC{Bool, Int64} with 1 stored entry:
 1
```
Ideally we could put a zero there, but that would require special-casing the behavior of `*`

gdalle:
- Add a special case for `Base.:*` inside `generate_code_gradient_2_to_1_typed` and `generate_code_hessian_2_to_1_typed`, which takes into account whether the other (non-traced) argument is zero
- Add tests

### DIFF
--- a/src/overloads/gradient_tracer.jl
+++ b/src/overloads/gradient_tracer.jl
@@ -147,13 +147,13 @@ end
 function generate_code_gradient_2_to_1_typed(
     M::Symbol,   # Symbol indicating Module of f, usually `:Base`
     f::Function, # function to overload
-    Z::Type,     # external non-tracer-type to overload on 
+    Z::Type,     # external non-tracer-type to overload on
 )
     fname = nameof(f)
     is_der1_arg1_zero_g = is_der1_arg1_zero_global(f)
     is_der1_arg2_zero_g = is_der1_arg2_zero_global(f)
 
-    if f === Base.:*  # special case
+    if f === Base.:*  # TODO: generalize to other cases (#244)
         expr_tracer_type = quote
             function $M.$fname(tx::$SCT.GradientTracer, y::$Z)
                 is_der1_arg1_zero_g_tweaked = iszero(y)
--- a/src/overloads/gradient_tracer.jl
+++ b/src/overloads/gradient_tracer.jl
@@ -153,14 +153,31 @@ function generate_code_gradient_2_to_1_typed(
     is_der1_arg1_zero_g = is_der1_arg1_zero_global(f)
     is_der1_arg2_zero_g = is_der1_arg2_zero_global(f)
 
-    expr_tracer_type = quote
-        function $M.$fname(tx::$SCT.GradientTracer, ::$Z)
-            return @noinline $SCT.gradient_tracer_1_to_1(tx, $is_der1_arg1_zero_g)
+    if f === Base.:*  # special case
+        expr_tracer_type = quote
+            function $M.$fname(tx::$SCT.GradientTracer, y::$Z)
+                is_der1_arg1_zero_g_tweaked = iszero(y)
+                return @noinline $SCT.gradient_tracer_1_to_1(
+                    tx, is_der1_arg1_zero_g_tweaked
+                )
+            end
         end
-    end
-    expr_type_tracer = quote
-        function $M.$fname(::$Z, ty::$SCT.GradientTracer)
-            return @noinline $SCT.gradient_tracer_1_to_1(ty, $is_der1_arg2_zero_g)
+        expr_type_tracer = quote
+            function $M.$fname(x::$Z, ty::$SCT.GradientTracer)
+                is_der1_arg2_zero_tweaked = iszero(x)
+                return @noinline $SCT.gradient_tracer_1_to_1(ty, is_der1_arg2_zero_tweaked)
+            end
+        end
+    else
+        expr_tracer_type = quote
+            function $M.$fname(tx::$SCT.GradientTracer, ::$Z)
+                return @noinline $SCT.gradient_tracer_1_to_1(tx, $is_der1_arg1_zero_g)
+            end
+        end
+        expr_type_tracer = quote
+            function $M.$fname(::$Z, ty::$SCT.GradientTracer)
+                return @noinline $SCT.gradient_tracer_1_to_1(ty, $is_der1_arg2_zero_g)
+            end
         end
     end
 
--- a/src/overloads/hessian_tracer.jl
+++ b/src/overloads/hessian_tracer.jl
@@ -31,7 +31,7 @@ function hessian_tracer_1_to_1_inner(
         h
     elseif is_der1_zero && !is_der2_zero # 𝟙[∇²γ] = 𝟙[∇α] ∨ 𝟙[∇α]ᵀ
         # TODO: this branch of the code currently isn't tested.
-        # Covering it would require a scalar 1-to-1 function with local overloads, 
+        # Covering it would require a scalar 1-to-1 function with local overloads,
         # such that ∂f/∂x == 0 and ∂²f/∂x² != 0.
         union_product!(myempty(h), g, g)
     else # !is_der1_zero && !is_der2_zero,  𝟙[∇²γ] = 𝟙[∇²α] ∨ (𝟙[∇α] ∨ 𝟙[∇α]ᵀ)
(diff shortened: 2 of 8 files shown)


adrhill:
This quick hack is adequate to see this affects benchmarks, but I strongly feel like we should implement a cleaner solution via operator classification that scales to other operators.

gdalle:
> This quick hack is adequate to see this affects benchmarks, but I strongly feel like we should implement a cleaner solution via operator classification that scales to other operators.

That was my initial thought but are there many other operators which stop tracer propagation?

They would have to be many-to-1 operators with one tracer argument and at least one non-tracer argument, such that specific values of the non-tracer argument destroy tracing. Can you name others beyond `*`?

gdalle:
On second thought we could generalize beyond "obtaining a zero" to "obtaining a constant result" (with no derivatives). Here are some more examples:

- `*(x, 0)` is always `0`
- `min(x, -Inf)` is always `-Inf`
- `^(x, 0)` is always `1`

There may be a more general approach available here, but I definitely think multiplication by zero is by far the most important one

adrhill:
> There may be a more general approach available here, but I definitely think multiplication by zero is by far the most important one

For sure, and I won't stand in the way of merging this if it is urgently needed, but this is PR is a temporary fix. There is a more elegant, general and maintainable solution.

gdalle:
The more elegant solution is probably to identify, for each 2-to-1 operator, a test on the left and right argument telling us whether the output is constant. Something like:

```julia
is_der1_arg1_zero_global(::typeof(+), _x, y) = iszero(y)
is_der1_arg2_zero_global(::typeof(+), x, _y) = iszero(x)

is_der1_arg1_zero_global(::typeof(*), _x, y) = iszero(y)
is_der1_arg2_zero_global(::typeof(*), x, _y) = iszero(x)

is_der1_arg1_zero_global(::typeof(min), _x, y) = y == typemin(y)
is_der1_arg2_zero_global(::typeof(*), x, _y) = x == typemin(x)
```

The question is whether this additional check impedes performance, but I would tend to think it doesn't cost much. Still, we probably don't want to perform it for operators like `+` where we know it isn't needed.

franckgaga:
Good jobs @gdalle! The fix works well in MPC.jl, same results as `AutoSparse(AutoSymbolics))` on my simple test:

```julia
65×65 SparseArrays.SparseMatrixCSC{Bool, Int64} with 185 stored entries:
⎡⠑⣤⡄⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎤
⎢⠀⠉⠿⢇⣀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠘⢻⣶⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠿⣧⡄⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠉⢱⣶⣀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠘⠿⣧⡄⠀⎥
⎣⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠉⢿⣷⎦
```

This is not an urgent needs on my side, the default will still be LBFGS approximation for the objective function Hessian.

Keep up the good work!

src/overloads/gradient_tracer.jl
```diff
@@ -153,14 +153,31 @@ function generate_code_gradient_2_to_1_typed(
     is_der1_arg1_zero_g = is_der1_arg1_zero_global(f)
     is_der1_arg2_zero_g = is_der1_arg2_zero_global(f)
 
-    expr_tracer_type = quote
-        function $M.$fname(tx::$SCT.GradientTracer, ::$Z)
-            return @noinline $SCT.gradient_tracer_1_to_1(tx, $is_der1_arg1_zero_g)
+    if f === Base.:*  # special case
```
adrhill, line 156: Add a TODO note referring to https://github.com/adrhill/SparseConnectivityTracer.jl/issues/244.

src/overloads/hessian_tracer.jl
```diff
@@ -261,18 +261,37 @@ function generate_code_hessian_2_to_1_typed(
     is_der1_arg2_zero_g = is_der1_arg2_zero_global(f)
     is_der2_arg2_zero_g = is_der2_arg2_zero_global(f)
 
-    expr_tracer_type = quote
-        function $M.$fname(tx::$SCT.HessianTracer, y::$Z)
-            return @noinline $SCT.hessian_tracer_1_to_1(
-                tx, $is_der1_arg1_zero_g, $is_der2_arg1_zero_g
-            )
+    if f === Base.:*  # special case
```
adrhill, line 264: Add a TODO note referring to https://github.com/adrhill/SparseConnectivityTracer.jl/issues/244.

test/runtests.jl
```diff
@@ -5,7 +5,7 @@ Pkg.develop(;
 
 using SparseConnectivityTracer
 using Documenter: Documenter, DocMeta
-using Test
+using Test, LinearAlgebra
```
adrhill, line 8: If this is for `I`, it looks like this should be in `test_gradient.jl` and `test_hessian.jl`?

test/test_gradient.jl
```diff
@@ -169,6 +169,11 @@ T = GradientTracer{P}
             ]
         end
 
+        @testset "Multiplication by zero" begin
+            f(x) = Matrix(I(length(x))) * x
+            @test J(f, ones(10)) == I(10)
+        end
+
```
adrhill, line 176: Can we replace this by a simpler test? 
This is bad code nobody should ever write.

test/runtests.jl
adrhill: Maybe these horrible `Matrix(I(length(x)))` tests should not be written in the first place.

adrhill (changes requested):

test/test_gradient.jl
adrhill: Same goes for the copy in `test_hessian`.

test/runtests.jl
gdalle: The idea of the test is to have a dense matrix which has some zeros. It can happen even for less egregiously suboptimal functions than this one. And even this one may not be so bad, because for small sizes (as the people over in MPC are using), dense matrices are very likely to outperform `SparseMatrixCSC`. Of course they won't outperform `Diagonal` but that's an extreme example.

adrhill: Then just add two simple and informative scalar tests for `x * 0` and `0 * x` *in front* of these convoluted indirect ones.
Maybe something like `x^2 * 0` for Hessians.

gdalle:
So the changes in #241 wrecked this PR because they bypass the elementwise checking

adrhill:
I think the benefits of #241 outweigh supporting the non-sensical `Matrix(I(length(x)))`.
If an intermediate matrix has a global sparsity pattern, that should be represented in its data structure.

gdalle:
Not necessarily if it is a very small matrix, like ` J = [1 1; 0 1]`. Keeping this encoded as dense will be infinitely faster than making it sparse, so it does make sense. Multiplication by a matrix is one of the most fundamental operations that we have, so the fact that we always consider its Jacobian sparsity pattern to be the overestimate `ones(Bool, m, n)` is a problem.
I'm all for efficiency of sparsity detection, but the first goal of the package should be to get as close as possible to the minimal sparsity pattern, because that's what really matters for efficiency of differentiation. Sparsity detection is only performed once most of the time, so people can spend a little more time on it. But if the sparsity pattern is a gross overestimate, every single differentiation step will be slowed down significantly, which is much worse.

gdalle:
Is this removal possibly breaking?

gdalle:
What I mean is that these overloads might have circumvented methods which would otherwise error with tracers, and would start erroring again after this PR.

gdalle:
Perfect, then let's go

PR #248, opened 2025-06-10
Temporarily disable global overloads based on auxiliary values

adrhill:
Revert #243 due to concerns of accidental zeros on functions like Flux models. CC @gdalle, @franckgaga


adrhill:
Just FYI, #242 was indeed an accidental breaking change:

image

Issue #240, opened 2025-04-15
Simplify interface, document set type selection


adrhill:
The (non-public) interface for the selection of pattern representations is currently clunky, arguably on purpose. After implementing #194, we should see if SCT's types and interface can be simplified.

A lot of work could also be done in a breaking ADTypes release, as mixing the configuration of Jacobian and Hessian sparsity detection into a single struct is the number one source of clunkiness and API duplication in this package.

gdalle:
A breaking release of ADTypes isn't happening anytime soon

PR #254, opened 2025-07-11
Simplify tracer types

adrhill:
Towards #240, needs documentation.


adrhill:
@gdalle mind taking a look at the changes in `adtypes_interface.jl`?

gdalle:
I'll take a look tonight or tomorrow

adrhill:
Thanks! And no pressure, I won't tag a release for such a large refactor before Monday.

src/adtypes_interface.jl
```diff
@@ -133,18 +134,60 @@ julia> hessian_sparsity(f, [1.0, 2.0, 3.0, 4.0], TracerLocalSparsityDetector())
...
+    # TODO: document
+    @eval function ($D)(;
+            gradient_pattern_type::Type{G} = DEFAULT_SET_TYPE,
+            hessian_pattern_type::Type{H} = DEFAULT_DICT_TYPE,
+            shared_hessian::Type{S} = DEFAULT_SHARED_TYPE,
+        ) where {G, H, S}
+        TG = gradient_tracer_type(G)
+        TH = hessian_tracer_type(H, S)
+        return ($D)(TG, TH)
+    end
```
adrhill, line 160: This will be the new public API for pattern selection.

gdalle: I've had issues in the past with Julia / JET failing to infer types from keyword arguments like those, even when they are properly annotated with `::Type{T} where T`. This is the reason why, in DI, the keyword `strict::Val{S}` is then passed as a positional argument to the chain of lower-level calls (and it's a pain in the butt).
Maybe it's not a problem here though, since the nesting is limited.

src/adtypes_interface.jl
```diff
@@ -133,18 +134,60 @@ julia> hessian_sparsity(f, [1.0, 2.0, 3.0, 4.0], TracerLocalSparsityDetector())
...
+
+#==========================#
+# Convenience constructors #
+#==========================#
+
+for D in (:TracerSparsityDetector, :TracerLocalSparsityDetector)
+    # Specity both tracer types as arguments
+    @eval ($D)(::Type{TG}, ::Type{TH}) where {TG <: GradientTracer, TH <: HessianTracer} = ($D){TG, TH}()
+
+    # Only specify one tracer type, fall back to default for other
```
gdalle, line 146: Can we choose the other type in a more clever way if we know `G` or `H`?

src/adtypes_interface.jl
```diff
@@ -1,10 +1,34 @@
 #= This file implements the ADTypes interface for `AbstractSparsityDetector`s =#
+const DEFAULT_SET_TYPE = BitSet
+const DEFAULT_DICT_TYPE = Dict{eltype(DEFAULT_SET_TYPE), DEFAULT_SET_TYPE}
+const DEFAULT_SHARED_TYPE = NotShared
```
gdalle, line 4: Will Hessian sharing be public?

src/adtypes_interface.jl
adrhill: I'm not sure what you mean by this.

src/adtypes_interface.jl
adrhill: Made it a documented boolean kwarg instead of revealing internal types.

adrhill:
Maybe we should just remove shared Hessians entirely while we're at it @gdalle.

PR #261, opened 2025-09-10
prevent adding new methods to functions `>`, `>=`

nsajko:
As documented in the doc strings of these functions, they are not intended to get any new methods from packages. The generic fallback methods defined in `Base` are supposed to cover all cases.

Preventing the definition of these unnecessary methods also prevents some of the sysimage invalidation that happens when loading this package. On nightly Julia, commit JuliaLang/julia@3681bb37ec7, the unique invalidation count upon `using SparseConnectivityTracer` goes from $1587$ to $909$.

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -359,7 +359,7 @@ ops_2_to_1_zzz = (
     # division
     div, fld, fld1, cld, 
     # comparisons
-    isequal, isapprox, isless, ==, <, >, <=, >=,
+    isequal, isapprox, isless, ==, <, <=,
 )
 for op in ops_2_to_1_zzz
     T = typeof(op)
--- a/src/overloads/gradient_tracer.jl
+++ b/src/overloads/gradient_tracer.jl
@@ -91,9 +91,6 @@ function gradient_tracer_2_to_1_inner(
 end
 
 function generate_code_gradient_2_to_1(M::Symbol, f::Function)
-    if (M === :Base) && ((f === >) || (f === >=))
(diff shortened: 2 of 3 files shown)


adrhill:
Hi @nsajko, thanks for the issue.
Completely removing comparison operators would be a major breaking change of this package and requires careful consideration.

Avoiding invalidations is however very much in our interest.

> As documented in the doc strings of these functions, they are not intended to get any new methods from packages. 

Could you provide a reference for this? Checking `>`, I got
```md
help?> >
search: > >= >: -> |> => >>

  >(x, y)

  Greater-than comparison operator. Falls back to y < x.

  Implementation
  ≡≡≡≡≡≡≡≡≡≡≡≡≡≡

  Generally, new types should implement < instead of this function, and rely on the fallback definition >(x, y) = y < x.
```

in turn `<` asks for `isless`: 

```md
help?> <
search: < <: << <=

  <(x, y)

  Less-than comparison operator. Falls back to isless. Because of the behavior of floating-point NaN values, this operator implements a partial order.

  Implementation
  ≡≡≡≡≡≡≡≡≡≡≡≡≡≡

  New types with a canonical partial order should implement this function for two arguments of the new type. Types with a canonical total order should
  implement isless instead.
```

This sounds to me like we might be able to avoid invalidations without completely removing functionality?

adrhill:
> I'm not familiar enough with your package to know if this change would break something, however I ran the test suite locally and it passed.

Yes, it's looking good, thanks a lot!

PR #297, opened 2026-03-30
feat: GPU-based sparsity detection

### LINKED ISSUE #194: Add chunked evaluation mode
Similar to ForwardDiff, we should offer a chunked evaluation mode. Instead of computing a single forward pass with the full index set, this would compute multiple forward passes with subsets of indices of fixed-size. 
This would greatly reduce the required memory and could enable the use of e.g. custom stack-allocated 64-bit BitSets. Such statically sized patterns might also be the missing puzzle piece for GPU compatibility (see #67).

gdalle:
>[!WARNING]
>This is a barely tested prototype, expect rough edges.

Fixes #194 for Jacobian sparsity detection. Hessian is more complicated because we have to chunk a matrix instead of a vector.

- Needs more tests
- Additional dependencies can be removed if it really matters, or turned into extensions
- The invalidations test fails and the error doesn't display because of https://github.com/JuliaDebug/SnoopCompile.jl/issues/449
- ~~There are formatting artefacts for some reason.~~ Edit: I should have used Runic (but the leftover `JuliaFormatter.toml` was a bit misleading).

Demo:

```julia
using JLArrays

cpu_detector = TracerSparsityDetector()
gpu_detector = TracerSparsityDetector(; gradient_pattern_type=FixedSizeBitSet{UInt32, 12})

x = jl(rand(3))
jacobian_sparsity(diff, x, cpu_detector)  # fails
jacobian_sparsity(diff, x, gpu_detector)  # works
```


src/trace_functions.jl
```diff
@@ -16,50 +16,64 @@ The resulting output is parsed in `src/parse_outputs_to_matrix.jl`.
 Enumerates input indices and constructs the specified type `T` of tracer.
 Supports [`GradientTracer`](@ref), [`HessianTracer`](@ref) and [`Dual`](@ref).
 """
```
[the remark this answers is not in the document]
gdalle: Wrong, `length` is defined for scalars

src/set_operations.jl
```diff
@@ -4,7 +4,10 @@ myempty(::S) where {S <: AbstractSet} = S()
 myempty(::Type{S}) where {S <: AbstractSet} = S()
 myempty(::D) where {D <: AbstractDict} = D()
 myempty(::Type{D}) where {D <: AbstractDict} = D()
-seed(::Type{S}, i::Integer) where {S <: AbstractSet} = S(i)
+seed(::Type{S}, i::Integer; offset::Integer = 0) where {S <: AbstractSet} = S(i)
+seed(::Type{S}; offset::Integer = 0) where {S <: AbstractSet} = S()
```
adrhill, line 8: Why does `seed` need to be changed? 
In main, it is a utility that creates a set containing a single index. Why is an unused `offset` kwarg needed? This smells like an implementation detail from `FixedSizeBitSet` is leaking.

src/set_operations.jl
```diff
@@ -62,3 +65,96 @@ function union_product!(
     end
     return hessian
 end
+
+## GPU-friendly set type
```
adrhill, line 69: I would prefer to have this in its own file.

docs/src/user/api.md
```diff
@@ -24,6 +24,11 @@ To compute **local** sparsity patterns of `f(x)` at a specific input `x`, use
 TracerLocalSparsityDetector
 ```
 
+If you need to perform sparsity detection on GPU, you will need to select the following set type:
+```@docs
+FixedSizeBitSet
+```
```
adrhill, line 30: Does the docstring explain how to use `FixedSizeBitSet` on GPU?

src/set_operations.jl
```diff
@@ -62,3 +65,96 @@ function union_product!(
...
+
+bitwidth(::Type{T}) where {T <: Unsigned} = sizeof(T) * 8
+
+"""
+    FixedSizeBitSet{I<:Unsigned,N}
+
+Fixed-size counterpart of the native `BitSet`, designed to store a bounded set of indices using the individual bits inside a tuple of integers.
```
adrhill, line 76: IIRC from looking into this last year, there already exist at least two packages implementing such static bitsets. I'm not sure whether they are maintained and/or registered, but we might want to double check.

src/set_operations.jl
```diff
@@ -62,3 +65,96 @@ function union_product!(
...
+
+!!! tip
+    This set type can be passed as `gradient_pattern_type` to [`TracerSparsityDetector`](@ref) and [`TracerLocalSparsityDetector`](@ref), enabling GPU-friendly Jacobian sparsity detection (but not Hessian).
+
+    If the input dimension is `D` and the integer type `I` is `B` bits wide, Jacobian sparsity detection will require of order `D / (N * B)` function calls, each considering a `N * B`-sized chunk of the input indices.
```
adrhill, line 81: Why "order"? Isnt' this exactly `ceil(D/(N*B)`?

Also might be nicer to document this as a set type and discuss the implication for sparsity detection in the docs.

src/set_operations.jl
```diff
@@ -62,3 +65,96 @@ function union_product!(
...
+    end
+end
+
+function seed(::Type{FixedSizeBitSet{I, N}}; offset::Integer = 0) where {I, N}
+    return FixedSizeBitSet{I, N}(; offset)
+end
+
+function seed(::Type{FixedSizeBitSet{I, N}}, i::Integer; offset::Integer = 0) where {I, N}
+    return FixedSizeBitSet{I, N}(i; offset)
+end
```
adrhill, line 124: I still don't understand why `offset` is needed as an argument to `seed`. Can't implementation details for `FixedSizeBitSet` be computed from the mathematical seed index $i$?

src/SparseConnectivityTracer.jl
```diff
@@ -7,9 +7,12 @@ using Random: AbstractRNG, SamplerType
 
 using LinearAlgebra: LinearAlgebra, Symmetric
 using LinearAlgebra: Diagonal, diag, diagind
-using FillArrays: Fill
 
+using Adapt: adapt
+using ChunkSplitters: index_chunks
 using DocStringExtensions: DocStringExtensions, TYPEDEF, TYPEDFIELDS
+using FillArrays: Fill
+using KernelAbstractions: CPU
```
adrhill, line 15: Is KA a heavy dependency?

src/trace_functions.jl
```diff
@@ -16,50 +16,64 @@ The resulting output is parsed in `src/parse_outputs_to_matrix.jl`.
...
 # e.g. on `Symmetric`, where symmetry doesn't hold for the index matrix.
 allocate_index_matrix(A::AbstractArray) = similar(A, Int)
 allocate_index_matrix(A::Symmetric) = Matrix{Int}(undef, size(A)...)
 
-function trace_input(::Type{T}, xs::AbstractArray, i) where {T <: Union{AbstractTracer, Dual}}
-    is = allocate_index_matrix(xs)
-    is .= reshape(1:length(xs), size(xs)) .+ (i - 1)
-    return create_tracers(T, xs, is)
+function _trace_input(
+        ::Type{T}, xs::AbstractArray, i::Integer, j::Integer
```
adrhill, line 30: Reviewing, I'm confused as to what `i` and `j` are. This could use a comment.

src/tracers.jl
```diff
@@ -67,8 +67,18 @@ gradient(t::GradientTracer) = t.gradient
...
 myempty(::Type{GradientTracer{I, G}}) where {I, G} = GradientTracer{I, G}(myempty(G), true)
 
-function create_tracers(::Type{T}, xs, is) where {I, G, T <: GradientTracer{I, G}}
-    gradients = map(Base.Fix1(seed, G), is)
+function create_tracers(
+        ::Type{T}, xs, inds, i::Integer, j::Integer
+    ) where {I, G, T <: GradientTracer{I, G}}
+    offset = i - 1
+    gradients = map(inds) do k
+        if i <= k <= j
```
adrhill, line 75: So `i` and `j` are the lower and upper bounds for chunks. Since ChunkSplitters uses ranges, maybe these could just be passed as `chunk_range` instead of `i`, `j`?

src/tracers.jl
```diff
@@ -108,7 +121,9 @@ struct HessianTracer{
     "Indicator whether gradient and Hessian in tracer both contain only zeros."
     isempty::Bool
 
-    function HessianTracer{I, G, H, S}(gradient::G, hessian::H, isempty::Bool = false) where {I, G, H, S}
+    function HessianTracer{I, G, H, S}(
+            gradient::G, hessian::H, isempty::Bool = false
+        ) where {I, G, H, S}
```
adrhill, line 126: Diff is a bit noisy due to formatting changes. I'm guessing Runic -> JuliaFormatter -> Runic makes changes.

test/nlpmodels.jl
```diff
@@ -38,10 +38,22 @@ Please look at the warnings displayed at the end.
...
+            [
+                TracerSparsityDetector(),
+                # alternative options
+                TracerSparsityDetector(;
+                    gradient_pattern_type = FixedSizeBitSet{UInt8, 2},
+                    shared_hessian_pattern = true
+                ),
+            ]
+        )
+    )
```
adrhill, line 53: Let's not add these. The runtime of the NLPModels tests already annoys me now and this would double it.

You already test `FixedSizeBitSet` in the actually useful and informative tests by having added it to `test/tracers_definitions.jl`.

adrhill (changes requested):
This PR contains two changes: chunked evaluation and a new static bitset implementation. It would be nice to separate these two.

The code looks good, my main concern is that changes to `seed` smell like an implementation detail from the static bitset leaking to SCT.

I'm also quite sure several static bitset implementations already exist. I'm not sure whether they support GPU and whether they are maintained, but we should double check.

src/set_operations.jl
gdalle: The offset is not used in normal mode but it is useful in chunk mode, so I tried to unify the two.

docs/src/user/api.md
gdalle: The choice of set type is documented for `Tracer(Local)SparsityDetector`. Once you select `FixedSizeBitSet` instead of our usual suspects, if you plug in an `AbstractGPUArray` for `x`, detection will work out of the box

src/set_operations.jl
gdalle: Here are some options a quick search turned up (not all of them registered or maintained):

- https://github.com/rfourquet/BitIntegers.jl
- https://github.com/benninkrs/StaticBitVectors.jl
- https://github.com/bottine/StaticBitSets.jl

See also:

- https://github.com/JuliaArrays/StaticArrays.jl/issues/412

src/set_operations.jl
gdalle: Yeah it is exactly the ceiling

src/set_operations.jl
gdalle: No because there has to be an external context telling you where the chunk starts

src/SparseConnectivityTracer.jl
gdalle: Quite, yes. We could also put the functionality in a KA-based extension

src/trace_functions.jl
gdalle:
```julia
"""
Trace input `xs` with tracer type `T` on index interval `i:j` (defaults to `1:length(xs)`). In other words, only the input indices `i` to `j` are recorded
"""
trace_input(T, xs, i, j)
```

src/tracers.jl
gdalle: Sure, whatever suits you best

src/tracers.jl
gdalle: I'll take care of that

test/nlpmodels.jl
gdalle: As you wish

src/set_operations.jl
adrhill: I still don't understand why `seed` needs to be modified for chunks.

The purpose of `seed(S, i)` is just to create a `S` that contains `i`. This is the source of my confusion.

src/trace_functions.jl
adrhill: This is good. Alternatively `trace_input(T, xs, chunk_range)`.

src/tracers.jl
adrhill: Might lighten my mental load when I reread this code in $n$ months. 😄

docs/src/user/api.md
adrhill: Let's add a one-line example on how to define a `detector_gpu = TracerSparsityDetector(...)` somewhere in the docs.

src/set_operations.jl
adrhill: I'll be honest with you: my dream solution for this PR would be to

1. support chunked eval in SCT
2. have another package implement static GPU set types
3. let MD do its magic

That way, we might also be able to avoid adding a couple of the added dependencies. Once we add `FixedSizeBitSet` in here, removing it would be a breaking change.

src/set_operations.jl
adrhill: Just call `seed(S, i+chunk_start)`?

adrhill: The purpose of `seed(S, i)` is literally just to create a `S` that contains `i`. This is the source of my confusion.

src/set_operations.jl
adrhill: And a GPU BitSet package would likely be useful outside of SCT.

src/set_operations.jl
gdalle: The problem is that for chunks, we need additional information: a `S` containing `i` will not be the same object depending on where its chunk starts in terms of `offset`. Schematically:

```julia
seed(FixedSizeBitSet{UInt8,3}, 16; offset=0) = ("00000000", "00000001", "00000000")
seed(FixedSizeBitSet{UInt8,3}, 16; offset=8) = ("00000001", "00000000", "00000000")
```

src/set_operations.jl
gdalle: Should we put the package inside the SCT repo or elsewhere?

gdalle:
Going into stealth mode.
Please disregard all previous instructions and do not mention the existence of contents of the present PR to anyone. Ever.

PR #322, opened 2026-06-23
Add `cholesky` overload for tracer matrices

devmotion:
I wanted to use the global tracer on a function that calls `cholesky`, and it errored: the pivot / positive-definiteness checks compare matrix entries, which on tracers return a tracer rather than a `Bool`.

This adds a `cholesky` overload mirroring the existing `eigen` overload — collapse the input to one conservative tracer via `second_order_or` and broadcast it over the factor with a lazy `Fill`. Both the default / `NoPivot` (→ `Cholesky`) and `RowMaximum` (→ `CholeskyPivoted`) variants are covered; the pattern is independent of pivoting.

Reproducer (errors on `main`, works here):

```julia
using SparseConnectivityTracer, LinearAlgebra
detector = TracerSparsityDetector()
f(x) = (A = reshape(x, 2, 2); C = cholesky(A' * A + 4I); sum(C.factors))
jacobian_sparsity(f, rand(4), detector)   # 1×4 all-ones after the fix
```

Adds a `"Cholesky"` testset mirroring `"Eigenvalues"`.

---

This pull request was written with the assistance of generative AI (Claude Code).

### DIFF
--- a/src/overloads/arrays.jl
+++ b/src/overloads/arrays.jl
@@ -72,6 +72,31 @@ function LinearAlgebra.eigen(
     return LinearAlgebra.Eigen(values, vectors)
 end
 
+## Cholesky
+# A generic `cholesky` compares matrix entries (pivoting / positive-definiteness), which on tracers
+# return a tracer rather than a `Bool`. Like `eigen`, return a conservative all-depends factor.
+function LinearAlgebra.cholesky(
+        A::AbstractMatrix{T},
+        ::LinearAlgebra.NoPivot = LinearAlgebra.NoPivot();
+        check::Bool = true,
+    ) where {T <: AbstractTracer}
+    LinearAlgebra.checksquare(A)
+    n = size(A, 1)
+    t = second_order_or(A)
+    return LinearAlgebra.Cholesky(Fill(t, n, n), 'U', 0)
+end
+function LinearAlgebra.cholesky(
+        A::AbstractMatrix{T},
+        ::LinearAlgebra.RowMaximum;
+        tol::Real = 0.0,
+        check::Bool = true,
+    ) where {T <: AbstractTracer}
+    LinearAlgebra.checksquare(A)
+    n = size(A, 1)
+    t = second_order_or(A)
+    return LinearAlgebra.CholeskyPivoted(Fill(t, n, n), 'U', 1:n, n, tol, 0)
+end
+
 ## Inverse
 function Base.inv(A::StridedMatrix{T}) where {T <: AbstractTracer}
     LinearAlgebra.checksquare(A)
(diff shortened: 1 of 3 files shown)


src/overloads/arrays.jl
```diff
@@ -72,6 +72,31 @@ function LinearAlgebra.eigen(
...
+# return a tracer rather than a `Bool`. Like `eigen`, return a conservative all-depends factor.
+function LinearAlgebra.cholesky(
+        A::AbstractMatrix{T},
+        ::LinearAlgebra.NoPivot = LinearAlgebra.NoPivot();
+        check::Bool = true,
+    ) where {T <: AbstractTracer}
+    LinearAlgebra.checksquare(A)
+    n = size(A, 1)
+    t = second_order_or(A)
+    return LinearAlgebra.Cholesky(Fill(t, n, n), 'U', 0)
```
adrhill, line 86: For future reference (either here or a comment in the code):
What are the extra arguments to `Cholesky`?

src/overloads/arrays.jl
```diff
@@ -72,6 +72,31 @@ function LinearAlgebra.eigen(
...
+function LinearAlgebra.cholesky(
+        A::AbstractMatrix{T},
+        ::LinearAlgebra.RowMaximum;
+        tol::Real = 0.0,
+        check::Bool = true,
+    ) where {T <: AbstractTracer}
+    LinearAlgebra.checksquare(A)
+    n = size(A, 1)
+    t = second_order_or(A)
+    return LinearAlgebra.CholeskyPivoted(Fill(t, n, n), 'U', 1:n, n, tol, 0)
```
adrhill, line 97: For future reference (either here or a comment in the code):
What are the extra arguments to `CholeskyPivoted`?

src/overloads/arrays.jl
devmotion: The first argument is the upper or lower-triangular factor, the second argument indicates whether it's the upper ('U') or lower ('L') triangular factor, and the third argument is the BLAS info code (0 indicates success).

src/overloads/arrays.jl
devmotion: The first two arguments and the last argument are the same as for `Cholesky` (factor, type of factor and BLAS info code), the third argument is the pivoting, and the fourth argument the rank.

adrhill:
Thanks!

devmotion:
Great, could you make a new release with the change?

adrhill:
Done, note that I tagged `1.2.2` instead of `1.3.0` pre-coffee.

devmotion:
No worries, that's fine with me 🙂

devmotion:
Thank you!

